PR #2, opened 2020-01-18
Project 15 - Sofie Nyblad

nyblad:
https://nyblad-todolist.netlify.com/

Fun project! And tricky to get the syntax and logic with the reducers and all Redux stuff, but I guess it will get clearer the more we code (hopefully).

**Things I'd like to have clarified or explained in more detail:**
- What is best practise of below alternatives?
```
removeAll: (state) => {
// return initialState 
state.items = []
}
```
- What is the difference between using props and this as Damien did in the recorded shopping-cart-codealong?
`export const CartItem = ({ product }) =>`
- Best practice to import custom fonts in React using styled components?
- My buttons with border-radius: 50% is not round in Safari, how to fix that?


code/src/components/Header.js
```diff
@@ -0,0 +1,83 @@
...
+      <WrapperRight>
+        <Text>{completedTasks.length}/{allTasks.length} done</Text>
+        <ClearButton onClick={handleClearTasks}>Clear all</ClearButton>
+      </WrapperRight>
+    </WrapperHeader>
+  )
+
+}
+
+const WrapperHeader = styled.section`
```
idlefingers, line 40: This is more of a preference thing, but generally I'd recommend having all the styles defined before the component itself. The way the JS compiles, it's not a problem, but technically, you should define things _before_ they're used in the file, so using these components in the `return` of your component and then defining them below is a bit backwards.

code/src/components/Header.js
```diff
@@ -0,0 +1,83 @@
...
+import moment from 'moment'
+import { useDispatch, useSelector } from 'react-redux'
+import { tasks } from 'reducers/tasks'
+
+export const Header = () => {
+
+  const dispatch = useDispatch()
+
+  // To get all tasks
+  const allTasks = useSelector(store => store.tasks.items)
```
idlefingers, line 12: Since you're only using the length of the array, I'd recommend returning that directly from the selector:

```
```

When you do it like you have, this `Header` component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the length of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.

code/src/components/TaskInput.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import styled from 'styled-components'
+import { useState } from 'react'
```
idlefingers: This import should be grouped with the main react import:

```
import React, { useState } from 'react'
```

code/src/components/TaskInput.js
```diff
@@ -0,0 +1,66 @@
...
+      <AddButton disabled={newTask.length === 0} onClick={handleAddTask}>+</AddButton>
+      <Input
+        type='text'
+        required
+        placeholder='Add task'
+        maxLength='100'
+        onChange={(event) => setNewTask(event.target.value)}
+        value={newTask}
+        // To add the task when hitting enter key
+        onKeyPress={(event) => { event.key === "Enter" && handleAddTask() }}
```
idlefingers, line 26: Nice solution!

If you wrapped the input in a `<form>` tag and put the `handleAddTask` function as the `onSubmit` function of the form instead, then this would be the default behavior and you wouldn't have to add this logic. Also, you'd be able to remove the `onClick` of your button (and give it `type="submit"` instead) and then everything would submit the form, which would then invoke the `handleAddTask` function.

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,41 @@
...
+const initialState = {
+  items: []
+}
+
+export const tasks = createSlice({
+
+  name: 'tasks',
+  initialState, //shorthand for initialState: initialState
+  reducers: {
+    // The actions for my store - but called reducers 🤷‍♀
```
idlefingers, line 13: 🤷‍♂

idlefingers (approved):
Super nice @nyblad! Great work!

> What is best practise of below alternatives?
> removeAll: (state) => {
> // return initialState 
> state.items = []
> }

Looks good to me?

> What is the difference between using props and this as Damien did in the recorded shopping-cart-codealong?
> `export const CartItem = ({ product }) =>`

This is [destructuring](https://simonsmith.io/destructuring-objects-as-function-parameters-in-es6). So if you had an object like `const person = { name: 'Damien' }` and you wanted to get the `name` value as a variable, you could do `const { name } = person` to kind of dig into the object.

You cal also do this when handling function arguments, so this is equivalent:

```jsx
const firstFunction = (props) => {
  console.log(props.name)
}

const secondFunction = ({ name }) => {
  console.log(name)
}

firstFunction({ name: 'Damien' })
secondFunction({ name: 'Damien' })
```

So it just saves you writing `props.whatever` in your component, making things a little more readable. But it also means you can have default values:

```jsx
const Greeting = ({ name = 'Bob' }) => (
  <h1>Hello {name}</h1>
)

<Greeting /> // would print <h1>Hello Bob</h1>
```

> Best practice to import custom fonts in React using styled components?

You can use [`createGlobalStyle`](https://styled-components.com/docs/api#createglobalstyle) to make a style which targets `body` and sets defaults for the page. `@import` is not recommended, though - https://styled-components.com/docs/faqs#note-regarding-css-import-and-createglobalstyle

> My buttons with border-radius: 50% is not round in Safari, how to fix that?

The checkbox circle? It looked fine to me in Safari, but if you have issues with things like this, it's often content or default styles messing with things. One approach could be to add a `<span>` inside the button or use a pseudo-selector like `::before` or `::after` to generate something 'clean' of default styles and content to work from.

Hope that cleared things up! Give me a shout if not :)

code/src/components/Header.js
nyblad: Will apply that in next project!

PR #3, opened 2020-01-19
project-todos by Agnes

Agnes94:
Link to my deployed project: https://agnesprojecttodos.netlify.com/

My reflections on how this week's project turned out: Really fun project ! I had a pretty bad week in general so I found it super hard to wrap my head around Redux even a little bit. On the positive side I felt like I understand React more than before so that feels good. I guess it needs more time and learning by doing.

Things I'd like to have clarified or explained in more detail: My delete button has a background on android devices for example, even though I took it away and looks fine on IOS so I don't really understand why is this.


code/src/components/EmptyList.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { useSelector } from "react-redux";
+import postIt from "assets/postIt.jpg";
+import "components/emptyList.css";
+
+export const EmptyList = () => {
+  const list = useSelector(state => state.tasks.items);
```
idlefingers, line 7: Since you're only using the length of the array, I'd recommend returning that directly from the selector:

```
```

When you do it like you have, the `EmptyList ` component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the length of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.

idlefingers (approved):
Great work @Agnes94! Really neat and tidy, and you're using redux really well! ⭐️ 

> My delete button has a background on android devices for example, even though I took it away and looks fine on IOS so I don't really understand why is this.

This is the default styles for your `<button>` element coming through, and different browsers behaving differently. A good approach here would be to just remove as much as possible from the defaults. I see you've removed the border in your css, but you could also remove the background and appearance (which is more browser defaults) to leave it completely up to your css to re-style the button (or just let the image do the work for you in this case):

```
.remove-button {
  border: none;
  appearance: none;
  background: none;
  padding: 0;
}
```

PR #4, opened 2020-01-19
Todos – Emmie Nilsson

eilsson:
Link to my deployed project:
https://redux-todo-app-emmie.netlify.com/

I feel like this project turned out ok. I'm most happy with the styling of the checkbox for each task. I probably could have made the components more reusable than I did. 

One thing I was thinking abou at the end of the project was if it could have been a good idea to use local storage to save state between sessions. I wasn't quite sure if it was or how I were to implement that.


code/src/components/Attribution.js
```diff
@@ -0,0 +1,25 @@
+import React from "react"
+
+import "./attribution.css"
+
+export const Attribution = () => {
+  return (
+    <details className="attributon-details">
+      <summary className="attribution-summary">Icon and image attribution</summary>
```
idlefingers, line 8: Nice! 🌟

code/src/components/ClearButton.js
```diff
@@ -0,0 +1,29 @@
...
+import { useDispatch, useSelector } from 'react-redux'
+import { todos } from 'reducers/todos'
+
+import "./clearButton.css"
+
+export const ClearButton = () => {
+
+  const dispatch = useDispatch()
+
+  const list = useSelector(state => state.todos.tasks)
```
idlefingers: Since you're only using the length of the array, I'd recommend returning that directly from the selector:

```
```

When you do it like you have, this `ClearButton ` component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the length of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.

code/src/components/EmptyListState.js
```diff
@@ -0,0 +1,23 @@
+import React from "react"
+import { useSelector } from "react-redux"
+import "./emptyListState.css"
+import listIllustration from "../icons/kingdom-list-is-empty.png"
+
+export const EmptyListState = () => {
+  const list = useSelector(state => state.todos.tasks)
```
idlefingers: Same here with returning exactly what you need from the selector.

code/src/reducers/todos.js
```diff
@@ -0,0 +1,26 @@
...
+      state.tasks.push({ id: Date.now(), text: action.payload, isCompleted: false })
+    },
+    removeTask: (state, action) => {
+      state.tasks = state.tasks.filter((task) => task.id !== action.payload)
+    },
+    toggleTaskStatus: (state, action) => {
+      const existingTask = state.tasks.find((task) => task.id === action.payload)
+
+      // If the task is already completed set isCompleted to false, otherwise to true to toggle task status
+      existingTask.isCompleted ? existingTask.isCompleted = false : existingTask.isCompleted = true
```
idlefingers: Doing assignment within a ternery like this is a little hard to read, I think. Since the `existingTask.isCompleted` condition returns a boolean, you could use `!` to invert it (so saying 'is not true') and use that directly and make this line shorter:

```js
existingTask.isCompleted = !existingTask.isCompleted
```

idlefingers (approved):
Really great work @eilsson! Really neat and tidy, and it looks like you have a good grasp of redux! ⭐️ 

> One thing I was thinking abou at the end of the project was if it could have been a good idea to use local storage to save state between sessions. I wasn't quite sure if it was or how I were to implement that.

This was a tough stretch goal! One solution is to use a middleware for redux - [redux-localstorage](https://www.npmjs.com/package/redux-localstorage). The readme for that package explains how to set it up when you're using 'vanilla' redux, but since we're using redux toolkit, you can add the middleware into the configureStore setup:

```js
// in App.js
import persistState from 'redux-localstorage'

  reducer,
  middleware: [
    persistState()
  ]
})
```

code/src/components/ClearButton.js
eilsson: Thank you, Damien. Wasn't sure that it was ok to do it like that. I'm making this change to this and empty list state.

code/src/reducers/todos.js
eilsson: Thank you! That is much better.

PR #5, opened 2020-01-19
To Do 

elol42:
Link to my deployed project:
https://jovial-shirley-7b2a15.netlify.com/

This week went okay, i didnt have that much time.. but i did one new thin in css, tooltips that i havent done before.. 


code/src/App.js
```diff
@@ -1,9 +1,35 @@
...
+
+import { Header } from 'components/Header'
+import { AddTask } from 'components/AddTask'
+import { TaskList } from 'components/TaskList'
+import { Counter } from 'components/Counter'
+import { ClearAll } from 'components/ClearAll'
+
+
+const reducer = combineReducers({
+  ToDoList: ToDoList.reducer
```
idlefingers, line 14: Reducers should have lower-case first letters (like a normal function) - only components and classes should start with uppercase letters. So:

```js
const reducer = combineReducers({
  toDoList: toDoList.reducer
})
```

The reducer should also export the slice as `toDoList` and then any component which uses the reducer would also need to be updated.

code/src/components/Task.js
```diff
@@ -0,0 +1,31 @@
...
+    const handleRemove = () => {
+        dispatch(ToDoList.actions.removeTask(props.task.id))
+    }
+
+    return (
+        <section className="listSection">
+             <li className={props.task.done ? 'checked' : 'unchecked'} >
+                <label className="checkboxLabel">
+                        <input type="checkbox" checked={props.task.done} onChange={handleCheckbox} />
+                            <span className="customCheckbox"></span>
```
idlefingers, line 22: Keep an eye on your indentation here! :)

code/src/components/TaskList.js
```diff
@@ -0,0 +1,17 @@
...
+
+export const TaskList = () => {
+    const tasks = useSelector((state) => state.ToDoList.tasks)
+    console.log(tasks)
+
+    return (
+        <div>
+        <ul>
+            {tasks.map((task) => ( <Task key={task.id} task={task} /> 
+            ))}
```
idlefingers, line 13: This should either be all on one line:

```js
{tasks.map((task) => <Task key={task.id} task={task} />)}
```

Body of the map function should be on a new line:

```js
{tasks.map((task) => (
  <Task key={task.id} task={task} /> 
))}
```

It's kind-of a mix of the two currently..

idlefingers (approved):
Nice work! Looks like you have a good feeling for redux! :) 

One thing - it looks like your editor is still set to use 4 spaces for indentation. I know I'm just being picky here, but it's good to get this set right - when you work in a dev team, they might want 2 spaces, 4 spaces, or tabs, and it's good to make sure you're ready to fit into it :)

PR #6, opened 2020-01-19
https://saras-todo.netlify.com/

Sara-Borgstrom:
Link to my deployed project: https://saras-todo.netlify.com/

Fun and some times hard. As allways I want more time but will give this project some more styling in the future. 

Styling on checkbox/radiobuttons.


code/src/components/ClearAll.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import { todoList } from '../reducers/todoList'
+import './clearall.css'
+
+export const ClearAll = () => {
+  const quantity = useSelector((state) => state.todoList.items)
+
+  const totalQuantity = quantity.length
```
idlefingers, line 9: Since you're only using the length of the array, I'd recommend returning that directly from the selector:

```
const numberOfTodos = useSelector(store => store. todoList.items.length)
```

When you do it like you have, this `ClearAll ` component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the length of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.

code/src/components/TaskBar.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import './taskbar.css'
+import { useSelector } from 'react-redux'
+
+
+export const TaskBar = () => {
+  const moment = require('moment');
+  const quantity = useSelector((state) => state.todoList.items)
+  const totalQuantity = quantity.length
```
idlefingers, line 9: Same here with just getting the length from the selector - it's a small optimisation, but good to think about! :)

code/src/components/TaskBar.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import './taskbar.css'
+import { useSelector } from 'react-redux'
+
+
+export const TaskBar = () => {
+  const moment = require('moment');
```
idlefingers, line 7: This import should go at the top of the file with the other imports. Also, `const moment = require('moment')` is a [commonjs import](https://blog.risingstack.com/node-js-at-scale-module-system-commonjs-require/).. It's the same as doing:

```js
import moment from 'moment'
```

code/src/components/TaskBar.js
```diff
@@ -0,0 +1,28 @@
...
+        <div className="container">
+          <p className="todo-text">TO DO</p>
+          <p className="date-box">{moment().format('MMM Do YY')}</p>
+        </div>
+      )}
+      {totalQuantity >= 1 && (
+        <div className="container">
+          <p className="todo-text">TO DO</p>
+          <p className="date-box">{moment().format('MMM Do YY')}</p>
+          <div className="amount">Tasks: {totalQuantity}</div>
```
idlefingers, line 22: It looks like this line is the only difference between these two blocks of code? I think it'd be much clearer to replace the two with something like this:

```js
<div className="container">
  <p className="todo-text">TO DO</p>
  <p className="date-box">{moment().format('MMM Do YY')}</p>
  {totalQuantity > 0 && <div className="amount">Tasks: {totalQuantity}</div>}
</div>
```

idlefingers (approved):
Great work! Super neat and tidy, and it feels like you're getting a really good grip of both react and redux - awesome! :) 

Regarding styling of inputs, I think you did great. It takes time to get used to doing these sort of fancy things and working around the html and default styles the browsers give you. I don't really have any good tips more than you've done already! One suggestion is, when you find an input on a site which you think is super nice, try to make a codepen or codesandbox and imitate it. Try to figure out how they did things using the inspector and reverse engineer it. 👍

PR #7, opened 2020-01-19
PR - Tiago Ferreira

tavferreira:
Link to my deployed project: https://pedantic-archimedes-f69b06.netlify.com

My reflections on how this week's project turned out: I've realized Redux was not entirely clear until I had to implement the reducer myself. I had some troubles at the beginning when dispatch the actions, but after the first code along all became clear. 

Things I'd like to have clarified or explained in more detail: I guess I would need more time to play with Redux to see if I fully grasp the concept. It's hard to say right now.


code/src/components/AddTask.js
```diff
@@ -0,0 +1,77 @@
...
+    setItem("");
+  };
+
+  return (
+    <TaskForm onSubmit={handleSubmit}>
+      <NewTask>
+        <AddNewTask
+          type="submit"
+          value="+"
+          active={item.length > 0 ? true : false}
```
idlefingers: Does this really need the ternary? 

This is like saying "If true return true else false". You could just use `active={item.length > 0}`

code/src/components/Header.js
```diff
@@ -0,0 +1,70 @@
+import React from "react";
+import { useSelector, useDispatch } from "react-redux";
+import moment from "moment";
```
idlefingers, line 3: Since you've installed the library 'react-moment', this import is actually importing the dependency 'moment' _from_ the 'react-moment' dependency. This can lead to unexpected things if you're not careful. I'd `npm uninstall react-moment` and replace it with the actual dependency; `npm install moment`, so the actual dependency is clear.

This is me being super picky, but it's good to be aware that importing dependencies of your dependencies is an npm-specific "feature" which is not super reliable. There are other package managers such as [yarn](https://yarnpkg.com/lang/en/) (my favourite! :) ) and [pnpm](https://pnpm.js.org/) where this import would actually fail since it's not a top-level dependency.

code/src/components/Header.js
```diff
@@ -0,0 +1,70 @@
...
+import styled from "styled-components";
+import { tasks } from "../reducers/tasks";
+
+export const Header = () => {
+  const taskStore = useSelector(store => store.tasks.tasks);
```
idlefingers: Since you're only using the length of the array, I'd recommend returning that directly from the selector:

```
const numberOfTasks = useSelector(store => store.tasks.tasks.length)
```

When you do it like you have, this `Header` component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the length of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.

code/src/components/List.js
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import { useSelector } from "react-redux";
+import { Task } from "components/Task";
+import styled from "styled-components";
+
+export const List = () => {
+  const taskStore = useSelector(store => store.tasks.tasks);
```
idlefingers, line 7: This one is a valid time to select the entire list rather than just the length of the array since you're using both further down

code/src/components/Task.js
```diff
@@ -0,0 +1,89 @@
...
+          type="checkbox"
+          checked={props.task.complete}
+          onChange={handleCheckboxClick}
+        />
+        <ItemInfo>
+          <ItemText completed={props.task.complete}>{props.task.text}</ItemText>
+          <ItemAdded completed={props.task.complete}>
+            {props.task.complete
+              ? "Completed " + moment(props.task.completed).fromNow()
+              : "Added " + moment(props.task.added).fromNow()}
```
idlefingers, line 27: Nice! ⭐️

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,61 @@
...
+import { createSlice } from "@reduxjs/toolkit";
+
+export const tasks = createSlice({
+  name: "tasks",
+  initialState: {
+    tasks: [
+      {
+        id: 1,
+        text: "Watch video on actions & reducers",
+        added: Date.now() - 604800000,
```
idlefingers: Complicated! Haha.

If this was in a project for a company, I'd want to see a comment explaining what this is to save me doing the maths. Or, since you have moment, you could just use that to make it more readable:

```js
added: moment().subtract(4, 'days')
```

idlefingers (approved):
Super nice @tavferreira! ⭐️ Looks like you have a good grasp of redux. Keep up the good work!

PR #8, opened 2020-01-19
Project 15 - Fredrik Tibbling

codebyfredrik:
https://peaceful-wiles-f65d36.netlify.com/

Yet another interesting project to work on. Using Redux really makes everything so much easier instead of passing props back and forth between components. 

Through out the project I have used the state in many different ways to control the UI. One feature that turned out great is the filter function. In the current implementation the user can filter and show all, completed, uncompleted or pinned todos. If the user selects the completed todos filter and then deletes all todos in that view, the filter will automatically revert back to filter show all, since there are no todos to show any more in the completed todos view. The logic is the same for all filters. If there are no todos to show for a particular filter, that filter is automatically disabled.

A user can create a new todo with any of the different filters selected. If the user for instance creates a new todo when the filter completed is selected, the newly created todo will not show up in the list. Therefore the filter will instantly revert back to the filter show all. The logic is the same for all filters.

Besides a clear all button I've also implemented a button to clear all completed todos (disabled if there are no completed todos).

Logic to be able to pin extra important todos to the top of the list has also been implemented. 

For the layout I have used Styled Components (with props) and CSS grid extensively to create a UI that adapts well to different screen sizes in both landscape and portrait mode. I'm really starting so see the real power of CSS grid.

I decided not to style the input checkbox in any way since the refreshed form controls in Chromium 81 looks really good by default. The design refresh is the result of an ongoing collaboration between Microsoft and Google

CSS breakpoints of interest:

- 500px
- 768px

**Things I'd like to have clarified or explained in more detail:** -


code/src/components/AddTodo.js
```diff
@@ -0,0 +1,81 @@
...
+
+  &:hover {
+    color: rgba(230, 0, 60, 1);
+    cursor: pointer;
+    transition: all 0.3s ease-in;
+  }
+`;
+
+export const AddTodo = () => {
+  const allTodos = useSelector(state => state.todos);
```
idlefingers: Since you're only using the length of the array, I'd recommend returning that directly from the selector:

```
const numberOfTodos = useSelector(store => store.todos.length)
```

When you do it like you have, this `AddTodo ` component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the length of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,81 @@
...
+  const [todoText, setTodoText] = useState('');
+  const dispatch = useDispatch();
+
+  const handleInputChange = e => {
+    setTodoText(e.target.value);
+  };
+
+  const handleSubmit = e => {
+    e.preventDefault();
+    if (!todoText.trim()) {
```
idlefingers, line 58: Nice! ⭐️

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,81 @@
...
+    dispatch(todos.actions.addTodo({ text: todoText }));
+    dispatch(visibilityFilter.actions.setVisibility({ filter: 'SHOW_ALL' }));
+    setTodoText('');
+  };
+
+  return (
+    <StyledAddTodo empty={allTodos.length === 0}>
+      <Form onSubmit={handleSubmit}>
+        <Button type="submit">
+          <i class="far fa-plus-square"></i>
```
idlefingers, line 70: I know it wasn't a requirement for this project, but don't forget to use aria-labels for this sort of stuff. This button isn't super accessible.

code/src/components/Checkbox.js
```diff
@@ -0,0 +1,56 @@
...
+
+const StyledCheckbox = styled.div`
+  display: inline-block;
+  width: 16px;
+  height: 16px;
+  background: ${props => (props.checked ? 'salmon' : 'papayawhip')}
+  border-radius: 3px;
+  transition: all 150ms;
+
+  ${HiddenCheckbox}:focus + & {
```
idlefingers, line 36: 😎

code/src/components/Filters.js
```diff
@@ -0,0 +1,105 @@
...
+  const countAll = useSelector(state => state.todos);
+  const countActive = useSelector(state =>
+    state.todos.filter(todo => todo.completed !== true)
+  );
+  const countCompleted = useSelector(state =>
+    state.todos.filter(todo => todo.completed === true)
+  );
+  const countPinned = useSelector(state =>
+    state.todos.filter(todo => todo.pinned === true)
+  );
```
idlefingers, line 55: Same with all these re: the point earlier about actually returning the count.

code/src/components/Footer.js
```diff
@@ -0,0 +1,65 @@
...
+    background-color: rgba(0, 117, 255, 1);
+    opacity: 1;
+    color: #ffffff;
+    transition: all 150ms ease-in;
+  }
+`;
+
+export const Footer = () => {
+  const dispatch = useDispatch();
+  const allTodos = useSelector(state => state.todos);
```
idlefingers, line 47: And here :D 

I'll stop commenting on each

code/src/components/Header.js
```diff
@@ -0,0 +1,28 @@
...
+
+const Title = styled.p``;
+
+const Date = styled.p`
+  color: rgba(0, 0, 0, 0.5);
+  font-size: 0.8rem;
+`;
+
+export const Header = () => {
+  let time = moment().format('MMMM Do, YYYY');
```
idlefingers, line 20: `const`!

code/src/components/Todo.js
```diff
@@ -0,0 +1,160 @@
+import React from 'react';
+import styled from 'styled-components';
+import { useDispatch } from 'react-redux';
+import { todos } from 'reducers/todos';
+
+const moment = require('moment');
```
idlefingers, line 6: This is a commonjs import. Since we have es6 all set up, you can write it like the other imports:

```js
import moment from 'moment'
```

code/src/components/Checkbox.js
```diff
@@ -0,0 +1,56 @@
...
+    box-shadow: 0 0 0 3px pink;
+  }
+
+  ${Icon} {
+    visibility: ${props => (props.checked ? 'visible' : 'hidden')}
+  }
+`;
+
+export const Checkbox = ({ checked }) => {
```
idlefingers, line 45: Is this actually being used?

code/src/components/Todo.js
```diff
@@ -0,0 +1,160 @@
...
+`;
+
+export const Todo = ({ todo }) => {
+  const dispatch = useDispatch();
+
+  return (
+    <StyledTodo>
+      <CheckboxWrapper>
+        <Checkbox
+          onChange={() => dispatch(todos.actions.toggleCompleted({ todo }))}
```
idlefingers, line 120: Nitpicking a bit here, but since this is reused so much, I think it'd be a little clearer if you made a function out of it and used that in all the `onChange` attrs. 

```js
const toggleCompleted = () => {
  dispatch(todos.actions.toggleCompleted({ todo }))
}

< Checkbox onChange={toggleCompleted} />
<TextWrapper onChange={toggleCompleted} />
//.. etc
```

... That way it's easier to see at a glance which ones are doing the same things.

code/src/components/Todos.js
```diff
@@ -0,0 +1,91 @@
...
+        dispatch(
+          visibilityFilter.actions.setVisibility({ filter: 'SHOW_ALL' })
+        );
+      }
+      break;
+    default:
+      filteredTodos = allTodos
+        .filter(todo => todo)
+        .sort((a, b) => b.pinned - a.pinned);
+  }
```
idlefingers, line 71: There's a whole lot going on here. You should aim to keep your components as dumb as possible. The side effects (changing filter if there's nothing to show) should happen in a reducer, and the all the filtering could be moved into fancier selectors, or potentially a custom hook. Something like this, perhaps:

```js
// In another file
const useFilteredTodosSelector = () => {
  const filter = useSelector(state => state.visibilityFilter)

  switch (filter) {
    case 'SHOW_ACTIVE':
      return useSelector(state => state.todos.filter(...).sort(...))
    case 'SHOW_COMPLETED':
      return useSelector(state => state.todos.filter(...).sort(...))
    ...
  }
}

// then, in this component:

const todos = useFilteredTodosSelector()
```

code/src/reducers/todos.js
```diff
@@ -0,0 +1,92 @@
...
+      completed: true
+    }
+  ],
+  reducers: {
+    addTodo: {
+      reducer(state, action) {
+        const { id, text, createdAt } = action.payload;
+        state.push({ id, text, createdAt, pinned: false, completed: false });
+      },
+      prepare({ text }) {
```
idlefingers, line 56: Nice! ⭐️

code/src/reducers/todos.js
```diff
@@ -0,0 +1,92 @@
...
+            createdAt: new Date().toISOString()
+          }
+        };
+      }
+    },
+    removeTodo: (state, action) => {
+      return state.filter(item => item.id !== action.payload.id);
+    },
+    clearAllTodos: (state, action) => {
+      return (state = []);
```
idlefingers: You can just do `return []` here

code/src/reducers/visibilityFilter.js
```diff
@@ -0,0 +1,11 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+export const visibilityFilter = createSlice({
+  name: 'visibilityFilter',
+  initialState: 'SHOW_ALL',
+  reducers: {
+    setVisibility: (state, action) => {
+      return (state = action.payload.filter);
```
idlefingers: This also doesn't need assignment. Just `return action.payload.filter`

idlefingers (approved):
Really nice @codebyfredrik! You've taken to redux like a fish to water. Looks like you have a really good grasp of it. Great work ⭐️

idlefingers:
One more comment just seeing your commit list now - try to write descriptive commit messages which describe what you worked on. It's great that you're committing super regularly, but having 20 commits with the same message is going to be a pain if you wanted to go back in history at some point (after finding a bug for example)

code/src/components/AddTodo.js
codebyfredrik: I agree that the component will be re-rendered every time I add/delete a task since the length of the array will change. But I don't see how the length of the array will change by just toggling a task's completion. The number of tasks remain the same the way I see it.

When I look at the code again I'm not even sure why I'm using the length value. It must have been for something, but I can't remember what. Looking inside the styled component the empty property is not used anywhere.

```
<StyledAddTodo empty={allTodos.length === 0}>
````

File: code/src/components/AddTodo.js (hunk 3 above)
Comment (codebyfredrik) [reply code/src/components/AddTodo.js:70]: I did some research and found this on Font Awesome's website: https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility

There are some pointers on how to handle semantic icons for screen readers properly.

File: code/src/components/Filters.js (hunk 5 above)
Comment (codebyfredrik) [reply code/src/components/Filters.js:55]: I agree.

With this modified code I don't get an array back, just the length from the useSelector hook:
```
 const countActive = useSelector(state => state.todos.filter(todo => todo.completed !== true).length);
```
Better?

code/src/components/Todo.js
codebyfredrik: Nitpicking is totally fine! 

It's very easy to drift away from the DRY principle. But I see your point here. I could restructure things a bit and make it more readable.

I don't have to pass the todo as an argument in the toggleCompleted function?

code/src/components/Todos.js
codebyfredrik: Looking at the code again I agree there is a lot going on. But I always want to improve myself so I will research and come up with a better solution. 

Despite that I'm happy I managed to implement the logic I had for my filter idea.

codebyfredrik:
> One more comment just seeing your commit list now - try to write descriptive commit messages which describe what you worked on. It's great that you're committing super regularly, but having 20 commits with the same message is going to be a pain if you wanted to go back in history at some point (after finding a bug for example)

When I'm working on a project like this I'm changing things back and forth quite a few times. That makes it hard to come up with descriptive commit messages after a while. It's just as hard as coming up with great variable or component names. It takes practice to become good at it.

But I will definitely work harder to iimprove this skill in future projects.

codebyfredrik:
Thanks again for all the great feedback.

PR #10, opened 2020-01-19
Project ToDo Frida J

hejfridafrej:
Link to my deployed project:
[https://frijon-todo.netlify.com/](https://frijon-todo.netlify.com/)

I think it gave me a good overview of how redux and the store works, but I still want to add some more features, and move the styling into styled components and do some conditional styling.

A little bit unsure of when the reducer needs an action and state, or just action or state as params. I felt like I just tried until it worked.


code/src/Components/List.js
```diff
@@ -0,0 +1,53 @@
...
+        type='button'
+        onClick={() => {
+          setShowTodo(false)
+          setShowDone(true)
+          setShowAll(false)
+        }}>
+        Show done
+      </button>
+​
+      <>{showAll && tasks.map(task => <Todo key={task.id} task={task} />)}</>
```
idlefingers, line 44: I don't think you need to use fragments here - `<>`. They're only needed when the `return` would return multiple elements instead of a single element. In this case, you're returning a single `<div>` so it's all good :)

code/src/Components/RemoveAll.js
```diff
@@ -0,0 +1,22 @@
...
+import { list } from 'reducers/list'
+
+
+
+export const RemoveAll = () => {
+    const dispatch = useDispatch()
+    
+    const removeAllTasks = () => {
+    dispatch(list.actions.removeAll())
+}
```
idlefingers, line 12: Keep an eye on your indentation here!

code/src/Components/RemoveAll.js
```diff
@@ -0,0 +1,22 @@
...
+    return(
+        <>
```
idlefingers, line 14: No need for a fragment here either - just returning the `<button>` is fine. To clarify:

```js
// Fine because it returns a single element:
return (
  <button>blah blah</button>
)

// Not fine - returns multiple elements so needs a fragment:
return (
  <button>blah blah</button>
  <button>blah blah</button>
  <button>blah blah</button>
  <button>blah blah</button>
)

// Fixed now since it just returns one element - a fragment:
return (
  <>
    <button>blah blah</button>
    <button>blah blah</button>
    <button>blah blah</button>
    <button>blah blah</button>
  </>
)
```

idlefingers (approved):
Nice work Frida! Looks like you have a good feel for redux - even if it doesn't feel like it just now.. It takes time! :) Keep it up!

> A little bit unsure of when the reducer needs an action and state, or just action or state as params. I felt like I just tried until it worked.

It looks like you've used it all correctly. So when you want to change something in some state, you need to use the state argument, and when you want to pass something into the reducer which isn't known to the reducer without an argument (such as the text of your todo item), then it needs to be passed in when you dispatch the action and it becomes `action.payload`. If all you want to do is replace the state entirely, then you dont need either:

```js
const myReducer = createSlice({
  name: 'example',
  initialState: {
    name: 'Damien'
  },
  reducers: {
    // In this action we want to change the `name` value in our 
    // state object, so we need the state argument and the action 
    // argument to get to the payload
    setName: (state, action) => {
      state.name = action.payload
    },

    // This action doesn't need to take the action argument since 
    // it doesn't care about the payload and does some hard-coded logic:
    setNameToBob = (state) => {
      state.name = 'Bob'
    },

    // This action completely replaces the state with a new object, so it 
    // doesn't need to take either state or action arguments:
    replaceState = () => {
      return { name: 'Joe' }
    },


    // But it really doesn't matter if you accept the arguments and don't 
    // use them. Redux will pass them in, but they're never used in this action:
    replaceStateButWhatevs: (state, action) => {
      return { name: 'Sue' }
    }
  }
})

// Example dispatches for these actions:

// changes state.name to 'Jane'
dispatch(myReducer.actions.setName('Jane'))

// Doesn't need any payload. I could pass one, but it'll just be ignored.
dispatch(myReducer.actions.setNameToBob())

// Same with these two. I could pass a payload, but it'll do nothing to change 
// what happens in the reducer
dispatch(myReducer.actions.replaceState())
dispatch(myReducer.actions.replaceStateButWhatevs())
```

PR #11, opened 2020-01-20
To-do app by Nasim Mahzoun

NasimMhn:
Link to my deployed project:
https://todo-app-by-nasim.netlify.com/
It turned out as a fun and enlightening project, even if there were many new things to take in. But I faced with some problems, but the hints and code alongs gave quite enough answers to my questions :)

- I would like to know, if I use the moment() function, to show the elapsed time, when a new task adds. you find it in my TasksList.js component.

 - I tried to overline the text, when a task complete, but it just didn't work at all. 
So in the "CheckBox" component it looks like this: 
 <label className="checkbox-label">
      <input
        type="checkbox"
        checked={props.task.complete}
        className={props.task.complete ? 'checked' : 'unchecked'}
        onChange={() => { handleCheckboxClick(props.task.id) }}>
      </input>
      <span className="checkmark" role="checkbox" aria-checked="false" tabIndex="0"></span>
    </label>
and for the styling: 
.checked .text {
  text-decoration: line-through;
} 
I would like to know, why is that?

Thank you very much.


code/src/components/TasksQuantity.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { useSelector } from 'react-redux'
+
+
+
+export const TasksQuantity = () => {
+  const allTasks = useSelector(state => state.tasks.items)
+  let completedTasks = allTasks.filter((item) => item.complete === true)
```
idlefingers, line 8: This should be `const` - it's not being reassigned!

code/src/components/TasksSummary.js
```diff
@@ -0,0 +1,21 @@
...
+//   return (
+//     <div className="tasks-summary-container">
+
+//       <div className="summary">
+//         <TasksQuantity />
+//         <span>You have {incompleteTasks.length} incomplete tasks</span>
+//       </div>
+//     </div>
+//   )
+// }
```
idlefingers, line 21: Don't forget to remove files which aren't used, just to tidy up the project

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,44 @@
...
+    items: taskData
+  },
+
+  reducers: {
+    addTask: (state, action) => {
+      state.items.push({ id: Date.now(), text: action.payload, complete: false, createdAt: moment().fromNow() })
+    },
+
+    toggleCheckbox: (state, action) => {
+      let foundItem = state.items.find((item) => item.id === action.payload)
```
idlefingers, line 26: This should be `const`

code/src/components/TasksList.js
```diff
@@ -0,0 +1,28 @@
...
+  // TODO - fetch all tasks from the store
+  const allTasks = useSelector((store) => store.tasks)
+
+  return (
+    <div className="tasks-container">
+      {allTasks.items.map((task) => (
+        <div key={task.id} className="task">
+          <CheckBox task={task} />
+          <div className="task-detail">
+            <div className="text"> {task.text} </div>
```
idlefingers, line 20: In answer to your question in the pull request about why the strikethrough css didn't work: If you want to write the css `.checked .text`, then you'd need to add the `checked` class to something which contains the .text element (such as the .task div on line 17 above. Or, a simpler way would be to add to the class on this line:

```
<div className={task.complete ? 'task checked' : 'task'}>{task.text}</div>
```

Then, in your css, you could select this text:

```css
.text.checked {
  text-decoration: line-through;
}
```

idlefingers (approved):
Great work Nasim! Looks like your getting a good grip on redux - keep it up! ⭐️

PR #13, opened 2020-01-20
Todo app - Aleksandra

AleksandraMuller:
Link to my deployed project: 
https://todo-app-aleksandra.netlify.com/

It was fun and challenging! I tried a lot of code before it worked out. If I had more time, I would make better styling choices and f.ex. make one styled component in order to follow DRY principle.


code/src/components/Date.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import moment from "moment";
+
+import styled from "styled-components";
+
+export const Date = () => {
+  const now = moment()
+    .format("LL")
+    .toString();
+  return <Container>Today: {now}</Container>;
```
JennieDalgren, line 10: Nice work with the date formatting and also that you made this into its own component

code/src/components/EmptyState.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+
+import styled from "styled-components";
+
+export const EmptyState = () => {
```
JennieDalgren, line 5: Nice to have an empty state component. You're getting the hang of the beauty of React and components now ☺️

code/src/components/Important.js
```diff
@@ -0,0 +1,29 @@
...
+    dispatch(tasks.actions.toggleImportant(id));
+  };
+  return (
+    <Button type="button" onClick={handleClick}>
+      {ifImportant === true && (
+        <ion-icon name="star" style={{ color: "orange" }}></ion-icon>
+      )}
+      {ifImportant === false && (
+        <ion-icon name="star-outline" style={{ color: "orange" }}></ion-icon>
+      )}
```
JennieDalgren: you could shorten this like:
`<ion-icon name={ifImportant ? "star" : "star-outline"} style={{ color: "orange" }}></ion-icon>`
since the name is the only thing that's different

code/src/components/Input.js
```diff
@@ -0,0 +1,76 @@
...
+      tasks.actions.addTask({
+        text: todoText,
+        id: taskId++
+      })
+    );
+  };
+
+  handleSubmit = e => {
+    e.preventDefault();
+  };
```
JennieDalgren: If you want to make it even more user friendly, so that you can just press enter when typed the todo instead of having to use the button, you could do line 15-21 inside the handleSubmit function instead. And then have the button onClick and the onSubmit calling that function.
```
 handleSubmit = e => {
    e.preventDefault();
    setTodoText("");
    dispatch(
      tasks.actions.addTask({
        text: todoText,
        id: taskId++
      })
    );
  };
```

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,57 @@
...
+      state.total -= 1;
+    },
+    showCompleted: state => {
+      state.taskData = state.taskData.filter(todo => todo.complete);
+      state.total = state.taskData.length;
+    },
+    showNotCompleted: state => {
+      state.taskData = state.taskData.filter(todo => !todo.complete);
+      state.total = state.taskData.length;
+    }
```
JennieDalgren: This doesn't really do what you want it to do I think.
When you have a list of 5 todos, for example. And then 2 of them are completed and 3 of them are not.
If you click to show only the completed ones, and then the ongoing ones you'll get no ongoing ones left. Since you made the taskData now be only the completed ones. This also messes up the total todos count.
You should modify the global state when you just want to filter the todos shown in different categories.
Does this make sense? If not, let me know!

JennieDalgren (approved):
Great job with this project and also with the styled components!

There are some minor things that you should look at to make the filtering work as you want it, now you're changing the whole todoDate array when you're filtering on either completed or ongoing tasks. So you'll loose some tasks there..

Check the comments in the code and let me know if you want me to explain more.

⭐️

code/src/components/Important.js
AleksandraMuller: True! Good point! : ) will fix that

code/src/components/Input.js
AleksandraMuller: 💪

code/src/reducers/tasks.js
AleksandraMuller: I think I know what you mean. I was sure it does what it suppose to, but now I know it is not. I will check this out 😆

PR #14, opened 2020-01-20
To-do-app Matilda Arvidsson

HIPPIEKICK:
Link to my deployed project: [https://matilda-arvidsson-project-todo-app.netlify.com/](https://matilda-arvidsson-project-todo-app.netlify.com/)

My reflections on how this week's project turned out: Fun project, and I feel like I'm finally _starting_ to understand Redux. Even though I followed the code alongs thoroughly, it feels like I understood what I did and not just copied. I had a hard time implementing the time stamp (and figuring out how to test if it worked since all new tasks said "a few seconds ago"). I hard coded the timestamp for the initialState (which I chose to keep), hope that's fine.

Things I'd like to have clarified or explained in more detail: I spent so much time trying to save my tasks in local storage. It seemed simple enough when I first read about it, but the more I googled, the more advanced it seemed. I tried this: [https://medium.com/@jrcreencia/persisting-redux-state-to-local-storage-f81eb0b90e7e]( https://medium.com/@jrcreencia/persisting-redux-state-to-local-storage-f81eb0b90e7e) but got TypeError: tasks.subscribe is not a function, and I couldn't get past it. I think you know "my level" more than I do, so please tell me if I'm in too deep water (that I'm not there yet) or point me in a direction so I can try again :)

Many thanks!


code/src/components/Header.js
```diff
@@ -0,0 +1,30 @@
...
+import { useSelector } from "react-redux"
+import { ClearButton } from "components/ClearButton"
+import moment from "moment"
+
+import "./header.css"
+
+export const Header = (state) => {
+  const items = useSelector((state) => state.tasks.items)
+
+  const completed = items.filter(item => item.checkedTask)
```
JennieDalgren, line 11: One minor little thing here is to actually use the selector to get only exactly what you want. So you just care about the length, so that's what you should return. I know I didn't show this in the codealong. (bad of me) 
`const items = useSelector((state) => state.tasks.items.length)`

When you do it like you have, this Header component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the **length** of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.

code/src/components/Item.js
```diff
@@ -0,0 +1,46 @@
...
+      </div>
+      <div className="date-and-remove-button">
+        <button
+          type="button"
+          onClick={handleRemoveButtonClick}
+          className="remove-task-button">
+          <span role="img" aria-label="remove-button">❌</span>
+        </button>
+        <p className="time-tag">
+          Added {moment(props.item.createdAt).fromNow()}
```
JennieDalgren, line 41: nice ⭐️

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,50 @@
...
+  ]
+}
+
+export const tasks = createSlice({
+  name: "tasks",
+  initialState,
+
+  reducers: {
+    //Here we will add actions
+    //Would make sense if this heading was actions instead lol
```
JennieDalgren, line 28: hehe nice comments

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+    removeAll: (state) => {
+      state.items = []
+    },
+    toggleCheckedTask: (state, action) => {
+      //find the right task and toggle the value of checkedTask
+      const checkedItem = state.items.find((item) => item.id === action.payload)
+
+      if (checkedItem) {
+        checkedItem.checkedTask = !checkedItem.checkedTask
```
JennieDalgren, line 45: ⭐️

JennieDalgren (approved):
Great job with the todo app!
It looks super.

_Regarding the local storage:_
This was a tough stretch goal! One solution is to use a middleware for redux - redux-localstorage. The readme for that package explains how to set it up when you're using 'vanilla' redux, but since we're using redux toolkit, you can add the middleware into the configureStore setup:
```
// in App.js
import persistState from 'redux-localstorage'

  reducer,
  middleware: [
    persistState()
  ]
})
```

Next week we'll learning how to create our own backend so then you could build a backend to this todo app and you'll have everything stored there!

PR #16, opened 2020-01-20
Todo project - Jenny Jacobsson

jennyjacobsson:
https://jenny-project-15-todo.netlify.com/

It was fun. Second week into Redux and things starts to get more understandable. I would like to have done some more of the stretch-goals but time ran out unfortunately.

**Things I'd like to have clarified or explained in more detail:**
Me and my team get this error in the browser sometimes but no one really understands it.

_Warning: A component is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component._

I understand that it has to do with the checkbox thing.  But I don't know how I should fix it?


code/src/components/AddTask.js
```diff
@@ -0,0 +1,29 @@
...
+
+export const AddTask = () => {
+  const [text, setText] = useState('')
+  const dispatch = useDispatch()
+
+  const handleSubmit = (event) => {
+    event.preventDefault()
+    dispatch(tasks.actions.addTodo(text))
+    setText('')
+  }
```
JennieDalgren, line 15: to make this even better you could check so that the text variable isn't empty before saving it as a todo? right now you could create empty todos.
```
const handleSubmit = (event) => {
  event.preventDefault()
  
  if (text) {
    dispatch(todo.actions.addItem(text))
    setText('')
  }
}
```

code/src/components/ClearButton.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { useDispatch, useSelector } from 'react-redux'
+import { tasks } from 'reducers/tasks'
+import 'css/buttons.css'
+
+export const ClearButton = () => {
+  const dispatch = useDispatch()
+  const allTasks = useSelector((state) => state.tasks)
```
JennieDalgren, line 8: Since you're only using the length of the array, I'd recommend returning that directly from the selector:

```
const allTasks = useSelector(state => state.tasks.length)
When you do it like you have, this ClearButton component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the length of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.
```

code/src/components/ToDoList.js
```diff
@@ -0,0 +1,24 @@
...
+  const allTasks = useSelector((store) => (store.tasks))
+
+  return (
+    <>
+      <h1>Todo List<span role="img" aria-label="Thinking cloud">  💬</span></h1>
+      {allTasks.map((task) => (
+        <Task key={task.id} task={task} />
+      ))}
+
+      {allTasks.length === 0
```
JennieDalgren, line 16: this could also be written `{!allTasks.length && ...}`

code/src/css/buttons.css
```diff
@@ -0,0 +1,46 @@
+.buttons {
+    display: flex;
+    margin:0 30px 0;
+    justify-content: space-evenly;
```
JennieDalgren, line 4: be consistent with spacing here.
`margin: 0 30px 0;`

JennieDalgren: And everywhere in the css 😉

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,31 @@
...
+
+    removeAll: () => {
+      return []
+    },
+
+    toggleCompleted: (state, action) => {
+      const foundTask = state.find((task) => task.id === action.payload)
+
+      if (foundTask) {
+        foundTask.complete = !foundTask.complete
```
JennieDalgren, line 26: ⭐️

JennieDalgren (approved):
Great job with this project!

I can't see that error that you mentioned in your comment but if you got that it has to do with that something, maybe the input value wasn't setup correctly or the text state was initially undefined instead of ' ' ? Let me know if you have that code somewhere and we can take a look?
Also, [this](https://stackoverflow.com/questions/56482774/uncontrolled-input-react-hooks) might help?

PR #17, opened 2020-01-20
Project todos, Michel Elmoznino Laufer

MichelLaufer:
Link to my deployed project:
https://michel-todo.netlify.com/

I think that both the structure of the project and the code-alongs were very good! I think that the "shopping cart" gave a good introduction to get started with this project and that the Wednesday and Friday code-alongs were both helpful in getting deeper into the project.

In terms of practicing Redux - how to write reducers and actions - I think that most of my questions were covered in the code-alongs. Maybe it would be nice to practice more styling using styled components and to get more inspo about designing todo's but as I understood the task the styling should be relatively "clean and simple" with the main focus being on Redux actions and reducers.


code/src/App.js
```diff
@@ -1,9 +1,27 @@
...
+
+const store = configureStore( { reducer } )
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="main-container">
+      <Provider store={store}>
+        <Head />
```
JennieDalgren, line 21: I think normally you would call a component like this for Header instead of Head. But that's up to you!

code/src/components/AddTask.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const AddTask = () => {
+  const [newTask, setNewTask] = useState('')
+  const dispatch = useDispatch()
+
+  const handleSubmit = (event) => {
+    event.preventDefault()
+    dispatch(tasks.actions.addTask(newTask))
+    setNewTask('')
+  }
```
JennieDalgren, line 13: one enhancement here could be to check if the input field is empty, and if it is don't add the task to the list.

code/src/components/Head.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import { tasks } from 'reducers/tasks'
+import moment from 'moment'
+
+export const Head = () => {
+  const dispatch = useDispatch()
+  const taskList = useSelector((store) => store.tasks.items)
```
JennieDalgren, line 8: Since you're only using the length of the array, I'd recommend returning that directly from the selector:
(I know I didn't show this in the codealong. My bad)

`const numberOfTasks = useSelector(store => store.tasks.items.length)`
When you do it like you have, this component will re-render whenever the array of tasks changes (for example toggling a task's completion), but if you return the length from the selector, it'll only re-render when the length of the array changes. Small optimisation, but good to keep in the back of your head if you're working on a bigger project where you want to avoid unnecessary re-renders.

code/src/components/TaskList.js
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <div>
+      <ul>
+        {taskList.map((task) => (
+          <Task key={task.id} task={task} timeCreated={task.timeCreated}/>
+        ))}
+      </ul>
+      <div className="uncompleted-tasks">
+        {countUncompleted.length} uncompleted task
+        {countUncompleted.length === 1 ? "" : "s"}
```
JennieDalgren, line 19: nice 😉

JennieDalgren (approved):
Great job with this project.

I think actions and reducers part will become more and more clear the more you use it.

Keep up the good work! ⭐️

PR #21, opened 2020-01-20
Todo Today

johannawaern:
Link to my deployed project:
https://todo-today.netlify.com/ 

Happy that it works, but so hard to grasp the concept of Redux. It's so much to keep in mind, my head feels like crows nest...

Redux ;). For me it's kind of hard to learn so many new things and just get like 20 minutes of video to explain it. I would love for a "Redux for dummies" video/code-along. The most simple project possible to build in Redux and talking about everything like I'm five years old. And a comment on every code-snippet in the code-along, like "this item.id refers to this, and this item is just my name so it could be gurka, and here is a function that does something and see what happens in my reducer...." Like soooo basic.


code/src/components/AddTodo.js
```diff
@@ -0,0 +1,28 @@
...
+
+export const AddTodo = () => {
+  const [todo, setTodo] = useState("")
+  const dispatch = useDispatch()
+
+  const handleSubmit = event => {
+    event.preventDefault()
+    dispatch(todoList.actions.addTodo(todo))
+    setTodo("")
+  }
```
JennieDalgren, line 13: to make this even better you could check so that the todo variable isn't empty before saving it as a todo? right now you could create empty todos.
```
const handleSubmit = (event) => {
  event.preventDefault()
  
  if (todo) {
    dispatch(todo.actions.addItem(todo))
    setTodo('') 
  }
}
```

code/src/components/CompletedItem.js
```diff
@@ -0,0 +1,30 @@
...
+    <li className='checked'>
+      <label className='container'>
+        <input type='checkbox' checked={true} onChange={handleCheckboxClick} />
+        <span class='checkmark'></span>
+      </label>
+      <div>{props.item.name}</div>
+      <button className='remove' type='button' onClick={handleRemoveClick}>
+        <span role='img' aria-label='remove todo'>
+          🗑️
+        </span>
```
JennieDalgren, line 26: ⭐️

code/src/components/CompletedList.js
```diff
@@ -0,0 +1,16 @@
...
+import { CompletedItem } from "components/CompletedItem"
+
+export const CompletedList = () => {
+  const todoItems = useSelector(state => state.todoList.completedItems)
+
+  return (
+    <ul>
+      {todoItems.length > 0 && <h2>Completed todo</h2>}
+      {todoItems.map(item => (
+        <CompletedItem key={item.id} item={item} />
```
JennieDalgren, line 12: This component is exactly the same as the ` <UncompletedItem />` right?
In that case, its better that you create one called **TodoItem** or similar that you can use in both lists.

code/src/components/UncompletedItem.js
```diff
@@ -0,0 +1,30 @@
+import React from "react"
```
JennieDalgren, line 1: duplicated component, more or less?
See my comment above

code/src/reducers/todoList.js
```diff
@@ -0,0 +1,34 @@
+import { createSlice } from "@reduxjs/toolkit"
+
+export const todoList = createSlice({
+  name: "todoList",
+  initialState: {
+    items: [],
+    completedItems: []
```
JennieDalgren, line 7: you could save some coding by only storing the items in one array and then have a property in each todo to be completed = true or false and filter the items array based on that.

JennieDalgren (approved):
Great job with this project. Also, it looks so nice and clean! ⭐️ 
Great design and UX.

Regarding Redux for dummies, there are a lot of resources out there. I'll try to dig up some good tutorials or articles to share with you.

One way of practicing this could also be to explain a code in a project for someone else. Like the quiz project or this todo project or something. You learn a lot by talking about your code!

Keep it up 🎉

PR #24, opened 2020-01-21
Todo project

horsefacegirl:
Link to my deployed project:
https://mystifying-pare-4f9c6f.netlify.com/

It was fun and I think I got a little bit better understanding of how Redux is structured. If I had the time I would have liked to do an animation when sorting the array and put a bit more effort on the styling.

How to think when debugging code and different approaches to do that.

### DIFF
--- a/code/src/components/AddTodo.js
+++ b/code/src/components/AddTodo.js
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 import { useDispatch } from 'react-redux'
 import { todoList } from 'reducers/todoList'
+import { Form } from 'components/styles'
 
 export const AddTodo = () => {
   const [todo, setTodo] = useState('')
@@ -13,7 +14,7 @@ export const AddTodo = () => {
   }
 
   return (
-    <form onSubmit={handleSubmit}>
+    <Form onSubmit={handleSubmit}>
       <label>
         +
         <input
@@ -23,12 +24,6 @@ export const AddTodo = () => {
         ></input>
       </label>
       <button type='submit'>Add</button>
-      {/* <button
-        type='button'
-        onClick={() => dispatch(todoList.actions.addTodo(item))}
-      >
-        Add To Do
-      </button> */}
-    </form>
+    </Form>
   )
 }
--- a/code/src/components/AddTodo.js
+++ b/code/src/components/AddTodo.js
@@ -1,9 +1,34 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { useDispatch } from 'react-redux'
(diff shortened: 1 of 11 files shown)


code/src/components/AddTodo.js
```diff
@@ -0,0 +1,40 @@
...
+        <ToggleAddTodo onClick={() => setShowInput(true)}>
+          + Add task
+        </ToggleAddTodo>
+      ) : (
+        <Form onSubmit={handleSubmit}>
+          <textarea
+            type="text"
+            value={todo}
+            onChange={event => setTodo(event.target.value)}
+          ></textarea>
```
idlefingers, line 30: In React, when a component or html element has no children, you can write it as a self-closing tag, so this and the button beneath it could be:

```
<textarea
  type="text"
  value={todo}
  onChange={event => setTodo(event.target.value)} />

<AddTodoButton
  disabled={todo.length === 0}
  type="submit"
  value="Add task" />
```

code/src/components/Header.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { useSelector } from "react-redux";
+import { HeaderWrapper } from "components/styles";
+
+export const Header = () => {
+  const now = new Date(Date.now());
+  const options = { dateStyle: "long" };
+  const date = new Intl.DateTimeFormat("sv-SE", options).format(now);
```
idlefingers, line 8: Nice!

code/src/components/Header.js
```diff
@@ -0,0 +1,26 @@
...
+
+  const incomplete = useSelector(store =>
+    store.todoList.items.reduce((total, item) => {
```
idlefingers, line 11: Nice use of `.reduce()`! 😎

code/src/components/TodoList.js
```diff
@@ -0,0 +1,15 @@
...
+import { TodoItem } from './TodoItem'
+import { List } from 'components/styles'
+
+
+export const TodoList = () => {
+  const todoItems = useSelector(state => state.todoList.items)
+  console.log(todoItems)
+  return (
+    <List>
+      {todoItems.slice().sort((a, b) => a.complete - b.complete).map(item => (<TodoItem key={item.id} item={item} />))}
```
idlefingers, line 12: I think this sorting should be done in the selector

idlefingers (approved):
Great work @horsefacegirl! Keep it up - looks like you're really getting a good grip on React and Redux. Awesome!

> How to think when debugging code and different approaches to do that.

That's a very broad question! Hard to answer when it's so broad - can you be a little more specific?

PR #26, opened 2020-01-21
https://questlog.netlify.com/

KarolCislowski:
Link to my deployed project:
https://questlog.netlify.com/

Very easy project, but unfortunately I have a heavy loaded schedule at the day job and can't spend extra time to polish it more. 


code/src/components/Form.js
```diff
@@ -0,0 +1,42 @@
...
+  const handleChangeCategory = (e) => setCategory(e.target.value)
+  const handleSubmit = (e) => {
+    e.preventDefault()
+    dispatch(quests.actions.createQuest({ text, category }))
+    setText('')
+  }
+  const handleDeleteAll = () => (dispatch(quests.actions.deleteAllQuests()))
+
+  return (
+    <>
```
idlefingers, line 22: No need for a fragment here. You only need fragments when you want to return multiple elements, but here there's just one `<form>` so you can return it directly.

code/src/components/Form.js
```diff
@@ -0,0 +1,42 @@
...
+      <form onSubmit={handleSubmit}>
+        <label htmlFor="newQuest">My next adventure will be:</label>
+        <input type="text" maxlength="20" value={text} required onChange={handleChangeText} id="newQuest" />
```
idlefingers, line 25: 20 character quest! Gotta get creative with my quest titles!

code/src/components/QuestLog.js
```diff
@@ -0,0 +1,45 @@
...
+
+  questsList.forEach((e, index) => {
+    if (e.category === 'Main') {
+      mainQuests.push(index)
+    } else if (e.category === 'Side') {
+      sideQuests.push(index)
+    } else {
+      miscellaneousQuests.push(index)
+    }
+  })
```
idlefingers, line 21: I think it would've been nicer to use `.filter` on your `questsLists` to find the items you're looking for:

```
const mainQuests = questsList.filter((quest) => quest.category === 'Main'))
const sideQuests = questsList.filter((quest) => quest.category === 'Side'))
const miscellaneousQuests = questsList.filter((quest) => !['Main', 'Side'].includes(quest.category)))
```

... Or just use 3 selectors and get what you want directly from redux.

code/src/components/QuestLog.js
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <main>
+      {(mainQuests.length > 0)
+        && <>
+          <h3>Main</h3>
+          {mainQuests.map((e) => (
+            <Quest key={`quest-${e}`} questIndex={e} />))}
```
idlefingers, line 29: When you have the entire quest object, why only pass the index? I think the `Quest` component could've just used the quest you pass it rather than looking into the store itself.

code/src/reducer/quests.js
```diff
@@ -0,0 +1,35 @@
...
+  categories
+}
+
+export const quests = createSlice({
+  name: 'quests',
+  initialState,
+  reducers: {
+    createQuest: (state, action) => {
+      const { text, category } = action.payload
+      state.questsList.push({
```
idlefingers, line 17: Don't forget the importance of IDs, or some form of unique keys in react. In your code, if you were to implement reordering of quests, react would have a hard time to see something has changed since you're relying entirely of the index of items in the array. It's much more solid to use unique IDs to track things since the array order can change

idlefingers (approved):
Nice work @KarolCislowski! Looks like you have a good grasp of react and redux.

I think you're doing great work, but one thing which I'd like you to push yourself to focus on is to not overcomplicate things - try to find the simplest route to a solution and remember - less is more! :)

PR #27, opened 2020-01-21
Project 15, To-Do Shoppinglist by Angelica Brodin

Angelicabrodin:
Link to my deployed project:
https://myshoppingtodos.netlify.com/

I have been very stressed these couple of weeks, ending an old job, starting a new one, so unfortunately the time for coding has been limited. With that in mind, I think it went pretty well. I still really struggle with redux though, and I kinda get the concept with dispatching actions, and the life cycle of redux, but when it comes to writing code, I have a hard time following why and when you are using the different hooks, props and functions, and what state that automatically follows into the component that you are mounting into App.js for example. And I am building my app completely by watching the code along, and I don't know if we should be able to build it completely by our self :D But as I said, I need to spend way more time reading and experimenting with the code to be able to better understand it I guess, its just practice. 
Once I get started though, it is really fun, and I think these kind of projects is really good! Also, I am not at all good in design, even though I love playing around with css, so I like the projects that have more of a design template, especially when it comes to putting it into the portfolio.

As stated before, it's the dispatching the actions, and the redux as a whole, that I found hard. 
Not something in particular, so I guess I just need more practice :)


code/src/components/ShoppingSummary.js
```diff
@@ -0,0 +1,28 @@
...
+    const items = useSelector(state => state.shopping.items)
+    const buyMoreItems = items.filter(item => item.needsMore === true)
+
+    const [showShoppingList, setShowShoppingList] = useState(false)
+
+    return (
+        <section className="shoppingSummary">
+            <h4>I have {items.length} items in my fridge</h4>
+            <h2>I need to buy more of {buyMoreItems.length} item
+                {buyMoreItems.length === 1 ? "." : "s."}
```
idlefingers, line 17: Nice with pluralization! I think it would be a little easier to understand with the entire word in there, but maybe that's just me being slow. The less you have to think when reading code to understand it, the better! Like this:

```js
{buyMoreItems.length === 1 ? "item" : "items"}
```

idlefingers (approved):
Really nice, clean code! Great work - keep it up. It takes time to get used to redux - it's not something which comes super naturally to most of us, since there's a lot of new concepts and ways of thinking. It took me quite a while to get used to it and confident to write things without having to reference other code, so don't worry - you're doing great! :)

PR #31, opened 2020-01-23
Late pull request - todos - Jola Chabik

ghost:
Link to my deployed project:

https://ucha3i-project-12.netlify.com/

I don't feel comfortable with Redux still. I did a to do with the example from the Redux documentations, and struggling with that  a lot. Finally, I feel like I understand the code, but I'm still not sure, if I'd be able to do it totally by myself.

I didn't put much effort into styling, because I'm already behind. I don't know how to make the checkboxes checked, and still looking for that. However, I decide to move on to next project, and do a pull request. I'd like to go back to this project and put it into my portfolio, when I'll have a bit more time. 

I don't really know, I don't get the Redux stuff, I don't feel that, but knowing myself, I think, I just need some time to grasp it, not exactly more explanation.

### DIFF
--- a/code/src/containers/FilterLink.js
+++ b/code/src/containers/FilterLink.js
@@ -1,0 +1,16 @@
+import { connect } from 'react-redux'
+import { setVisibilityFilter } from '../actions'
+import Link from '../components/Link'
+
+const mapStateToProps = (state, ownProps) => ({
+  active: ownProps.filter === state.visibilityFilter
+})
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  onClick: () => dispatch(setVisibilityFilter(ownProps.filter))
+})
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(Link)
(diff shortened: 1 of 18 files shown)


code/src/containers/FilterLink.js
```diff
@@ -0,0 +1,16 @@
...
+import { setVisibilityFilter } from '../actions'
+import Link from '../components/Link'
+
+const mapStateToProps = (state, ownProps) => ({
+  active: ownProps.filter === state.visibilityFilter
+})
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  onClick: () => dispatch(setVisibilityFilter(ownProps.filter))
+})
```
JennieDalgren, line 11: mapStateToProps and mapDispatchToProps is from before we had React Hooks.
Have a[ read here](https://blog.bitsrc.io/using-react-redux-hooks-97654aff01e4) regarding this

JennieDalgren (approved):
Remember, we're using React Hooks :)

PR #33, opened 2020-01-25
todo app code complete + styling

Lilivet:
Link to my deployed project:
https://todos-redux-ib.netlify.com/

I used styled components and it is easy, but not as easy as normal CSS. 

I think, I don´t get the key yet. I did not use it in the form first, but then when I made a production site, I saw that the app did not work in Safari. You could type only one letter at the time. I added a key in the Form component and re-build it and it started working properly...


code/src/App.js
```diff
@@ -1,9 +1,23 @@
...
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <Provider store={store}>
+      <Header />
+      <AddTodosForm />
+      <TodoList />
```
JennieDalgren, line 20: good structure with components.

code/src/components/AddTodosForm.js
```diff
@@ -0,0 +1,62 @@
...
+import styled from 'styled-components'
+
+const FormWrapper = styled.div`
+border-bottom: 2px solid #d1d1e0;
+height: 80px;
+background-color: #F1F9F9;
+padding: 0 5% 0 3%;
+display: flex;
+flex-direction: row;
+align-items: center;
```
JennieDalgren, line 13: indentation

code/src/components/AddTodosForm.js
```diff
@@ -0,0 +1,62 @@
...
+background: transparent;
+`;
+export const AddTodosForm = () => {
+    const [task, setTask] = useState("")
+    const dispatch = useDispatch()
+
+    const handleSubmit = (event) => {
+        event.preventDefault()
+        dispatch(todos.actions.addTask(task))
+        setTask('')
```
JennieDalgren, line 44: To make this even better, think about handling an empty input. Right now it's possible to save a blank todo.
Maybe add something like:
`if (!task) return` 
to make sure to return from the function if there's no task written

code/src/components/Header.js
```diff
@@ -0,0 +1,57 @@
...
+                <RegularText>{moment().format("MMM Do")}</RegularText>
+            </WrapperLeft>
+            <WrapperRight>
+                <RegularText>{tasks.length} Tasks</RegularText>
+                <ClearButton />
+            </WrapperRight>
+        </Wrapper>)
+}
+
+const Wrapper = styled.section`
```
JennieDalgren, line 23: try to stick to having the styled components on top of the file like you did above

code/src/components/Task.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react'
+import { useDispatch } from 'react-redux'
+import { todos } from '../reducers/todos'
+import styled from 'styled-components'
+import { Trash } from '../icons/Trash'
+
+export const Task = (props) => {
+    const dispatch = useDispatch()
+    const handleCheckboxClick = () => {
```
JennieDalgren, line 9: here you coudl destructure the props to not have to write props.item everywhere.
Either like this:
`const {id, name, complete} = props.item`
or when creating the Task component:
`export const Task = ( {id, name, complete} ) => { ... `

code/src/reducers/todos.js
```diff
@@ -0,0 +1,46 @@
...
+        // current state and action that it is going to be performed (state, action)
+        toggleCompleted: (state, action) => {
+            console.log(action.payload)
+            // need to find tasks like "Create a todo app"
+            const foundItem = state.items.find((item) => item.id === action.payload)
+            console.log('found', JSON.stringify(foundItem))
+            // and toggle its state under "complete"
+            if (foundItem) {
+                foundItem.complete = !foundItem.complete
+            }
```
JennieDalgren, line 42: good job with comments here.

JennieDalgren (approved):
Good job with the project!

You have a good grasp of React and Redux and also styled components!
Remember indentation in styled components.

PR #40, opened 2020-04-20
Hanna To do App (review wanted)

Ruterryter:
Link to my deployed project:
https://nostalgic-todo.netlify.app/

It was a fun project but a bit tricky as always. I was doing it mostly after Vans code along to really understand what i was doing. 

My redux devtools did not work, it said store not installed and I saw that someone else had the same issue. Tried to use configuretStore instead of createStore like Van did but did not get that to work so would like to know why and when we should use one over the other and why that had an impact on dev tools.


code/src/pages/Home.js
```diff
@@ -0,0 +1,40 @@
...
+import { Header } from "components/Header"
+import { CustomCheckbox } from 'components/CustomCheckbox';
+
+// Tell redux about our reducers
+const reducer = combineReducers({ todos: todos.reducer });
+
+// // Retrieve the existing state from localstorage if it exists
+const persistedState = localStorage.getItem("reduxState")
+  ? JSON.parse(localStorage.getItem("reduxState"))
+  : {};
```
JennieDalgren, line 15: Great job trying out Local storage.

code/src/index.css
```diff
@@ -5,9 +5,134 @@ body {
...
+    20% { transform: translate(-3px, 0px) rotate(1deg); }
+    30% { transform: translate(3px, 2px) rotate(0deg); }
+    40% { transform: translate(1px, -1px) rotate(1deg); }
+    50% { transform: translate(-1px, 2px) rotate(-1deg); }
+    60% { transform: translate(-3px, 1px) rotate(0deg); }
+    70% { transform: translate(3px, 1px) rotate(-1deg); }
+    80% { transform: translate(-1px, -1px) rotate(1deg); }
+    90% { transform: translate(1px, 2px) rotate(0deg); }
+    100% { transform: translate(1px, -2px) rotate(-1deg); }
+  }
```
JennieDalgren, line 132: nice little animation on the trash can

code/src/pages/Home.js
```diff
@@ -0,0 +1,40 @@
...
+
+// Create the store using our reducers and the retrieved state
+const store = createStore(
```
JennieDalgren, line 18: Stick to `configureStore` instead of `createStore`, since we're using Redux Toolkit and not the whole Redux library.
[https://redux-toolkit.js.org/api/configureStore](https://redux-toolkit.js.org/api/configureStore)

code/src/pages/Home.js
```diff
@@ -0,0 +1,40 @@
...
+  reducer,
+  persistedState,
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
```
JennieDalgren, line 21: I don't think you need this line if you change to configureStore

code/src/index.css
```diff
@@ -5,9 +5,134 @@ body {
...
+  font-size: 16px;
+  color: #1e3756;
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  margin-left: 7px;
+}
+
+.todo-item.done {
+text-decoration: line-through;
```
JennieDalgren: indentation 😉

code/src/components/TodoSummary.js
```diff
@@ -0,0 +1,18 @@
+import React, { useEffect, useState } from "react";
+import { TodoItem } from "./TodoItem.js";
+import { TodoInput } from "./TodoInput.js";
+import { useSelector } from "react-redux";
+
+export const TodoSummary = () => {
+  const list = useSelector(store => store.todos.list);
```
JennieDalgren, line 7: you are only interested in the items array, so better to get only that array from the store instead of the whole object.
```
const items = useSelector(store => store.todos.list.items);
```

code/src/components/TodoList.js
```diff
@@ -0,0 +1,21 @@
+import React, { useEffect, useState } from "react";
+import { TodoItem } from "./TodoItem.js";
+import { TodoInput } from "./TodoInput.js";
+import { TodoSummary } from "./TodoSummary.js";
+import { useSelector } from "react-redux";
+import { RemoveAll } from './RemoveAll'
+
+export const TodoList = () => {
+  const list = useSelector((store) => store.todos.list);
```
JennieDalgren, line 9: you are only interested in the items array, so better to get only that array from the store instead of the whole object.
```
const items = useSelector(store => store.todos.list.items);
```

code/src/components/TodoList.js
```diff
@@ -0,0 +1,21 @@
...
+
+  return (
+    <section className="todo-list">
+      <TodoInput />
+      {list.items.map((item, index) => (
+        <TodoItem item={item} itemIndex={index} />
```
JennieDalgren: Don't forget to add the key prop to the child here
`<TodoItem key={index} item={item} itemIndex={index} />`

code/src/components/TodoItem.js
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <div className={`todo-item ${item.done ? "done" : ""}`}>
+      <CustomCheckbox
+        isChecked={item.done ? "done" : ""}
+        onChangeHandler={handleCheckboxClick}
+      />
+      <span className="todo-item-description"> {item.description} </span>
+      <div className="created-at-time"> {moment(item.date).fromNow()}</div>
+      <div className="todo-item-remove" onClick={onRemoveClicked}>
+        <svg className="trash"
```
JennieDalgren, line 40: I would suggest to make this svg into its own icon component. It'll make it reusable and will make this TodoItem component a bit cleaner and easier to read.

code/src/components/TodoItem.js
```diff
@@ -0,0 +1,54 @@
...
+  const item = useSelector((store) => store.todos.list.items[itemIndex]);
+  const dispatch = useDispatch();
+
+  const handleCheckboxClick = (e) => {
+    dispatch(todos.actions.setDone({
+      itemIndex: itemIndex,
+      done: !item.done,
+    })
+    );
+  };
```
JennieDalgren: Heads up!
When you're using the index to indicate what item should be checked or not, there's a bug appearing.
Try to add a bunch of todos, check the first one but not the second one, the todo with index 0 will be checked. Then remove the first one.
You'll see that now the second one will be checked instead. This is because that one now gets the "new" index 0.

JennieDalgren: To fix this, try to use an `id` instead of an index to deal with the done state.

code/src/reducers/todos.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+};
+
+export const todos = createSlice({
+  name: "todos",
+  initialState: initialState,
+  reducers: {
+    addTodo: (state, action) => {
+      const { itemInfo } = action.payload;
+      state.list.items.push(itemInfo);
```
JennieDalgren: Also add an id to your todos to be able to control them in a better way.
You can use a timestamp as you id, then you know they will always be unique.

code/src/components/TodoItem.js
```diff
@@ -0,0 +1,54 @@
...
+
+  const onRemoveClicked = (e) => {
+    dispatch(
+      todos.actions.removeTodo({
+        itemIndex,
```
JennieDalgren: again, here try to use the id instead of index when removing. It's safer.

code/src/components/TodoInput.js
```diff
@@ -0,0 +1,48 @@
...
+import { useDispatch } from "react-redux";
+import { todos } from "../reducers/todos.js";
+
+
+export const TodoInput = ({ listId }) => {
+  const dispatch = useDispatch();
+  const [inputValue, setInputValue] = useState("");
+
+  const handleOnSubmit = e => {
+    // Prevent page reload
```
JennieDalgren, line 11: right now you can save an empty todo. to prevent this, check if the input is empty and if it is, don't dispatch the addTodo action. 
`if (!inputValue) return`
Another solution is to disable the add button until the input is filled in.

code/src/components/Header.js
```diff
@@ -0,0 +1,34 @@
...
+font-weight: 700;
+font-family: 'niveau-grotesk', sans-serif ;
+font-style: italic;
+`
+const StyledBackground = styled.div`
+background-color:#f18666;
+height: 50px;
+width: 100%;
+border-bottom-right-radius: 55% 120%;
+`
```
JennieDalgren: indentation

code/src/components/CustomCheckbox.js
```diff
@@ -0,0 +1,69 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+
```
JennieDalgren, line 3: nice work with a custom checkbox! ⭐

JennieDalgren (approved):
Great job with this project! 

I've answered you question about redux devtools in the code further down.
But mainly it has to do with Redux Toolkit vs Redux and when to use createStore or configureStore. Let me know if you want me to explain it further.

Keep up the good work! 💪

code/src/reducers/todos.js
Ruterryter: Hi, I tried to change this and add an id to the todos but cant get it to work. Could you give me some pointers how to change from index to id in the different files?

PR #43, opened 2020-04-22
review wanted

kajsabootin:
Link to my deployed project:

https://kajsabootin-todos.netlify.app/

I was struggling so much to style my checkboxes. It was really hard, and I look at many tutorials. And now is it the thing I am most proud of in this project.

I really appreciate Vans code along from last week where he did a todo.


code/src/App.js
```diff
@@ -1,9 +1,10 @@
-import React from 'react'
+import React from 'react';
+import {Home} from './pages/Home.js';
 
 export const App = () => {
   return (
     <div>
-      Find me in src/app.js!
+      <Home />
```
JennieDalgren: No need to wrap `<Home />` in a div here.

code/src/components/Header.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import moment from 'moment'
+import { todos } from "reducers/todos";
+import { useSelector, useDispatch } from "react-redux";
+
+export const Header = () => {
+  const list = useSelector((store) => store.todos.list);
+  const doneList = list.items.filter((item) => item.done);
```
JennieDalgren, line 8: if you're just interested in the length of this list, you could store only that in doneList variable.
 `const doneList = list.items.filter((item) => item.done).length`

code/src/components/Header.js
```diff
@@ -0,0 +1,28 @@
...
+  const dispatch = useDispatch();
+  
+  const onRemoveClicked = (event) => {
+    dispatch(todos.actions.removeAllTodos());
+  };
+
+  return (
+    <div className="header-container">
```
JennieDalgren, line 16: Just a minor thing but remember to still use semantic html elements, so maybe not divs everywhere but use sections, articles and so on.

code/src/components/TodoInput.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // E står för Event
+  const handleOnSubmit = e => {
+    // Prevent page reload
+    e.preventDefault();
+    //console.log(inputValue)
+  
+
+  // This is where we will Dispatch the action to save new todo item
+    dispatch(
```
JennieDalgren: nice work with comments to explain the code. But remove the console log 😉

JennieDalgren: Also, you might want to check that the inputValue is not an empty string before saving the todo. Right now you can get "empty" todos in your list.
For example:
```
const handleOnSubmit = e => {
    e.preventDefault();
    
    if (!inputValue) return // this will just stop the handleOnSubmit function from going further

    dispatch(todos.actions.addTodo({
      description: inputValue,
      done: false,
    })
  );
  setInputValue("");
};
```

code/src/components/TodoItem.js
```diff
@@ -0,0 +1,49 @@
...
+      todos.actions.removeTodo({
+        itemIndex,
+      })
+    );
+  };
+
+  const handleOnChange = (event) => {
+    dispatch(
+      todos.actions.setDone({
+        itemIndex: itemIndex,
```
JennieDalgren, line 21: since the key and value are the same its enough to just write `itemIndex` instead of `itemIndex: itemIndex`

Also, be aware that if you manipulate your todolist in a way so that the indexes changes of the todos, you might be changing or deleting the wrong todos here... Try to use an id for each todo instead of index.

code/src/components/TodoItem.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { useSelector, useDispatch } from "react-redux";
+import { todos } from "reducers/todos";
+
+
+export const TodoItem = ({itemIndex}) => {
+  const item = useSelector((store) => store.todos.list.items[itemIndex]);
```
JennieDalgren, line 7: you could also just pass in the whole item as a prop to this component instead of fetching it again from the global store...

code/src/components/TodoList.js
```diff
@@ -0,0 +1,16 @@
...
+import { TodoInput } from "./TodoInput";
+import { TodoItem } from "./TodoItem";
+
+export const TodoList = () => {
+  const list = useSelector(store => store.todos.list);
+  return (
+    <div className="todo-list">   
+      <TodoInput />
+      {list.items.map((item, index) => (
+      <TodoItem key={index} itemIndex={index}></TodoItem>
```
JennieDalgren, line 12: you could pass along the whole item here instead of just the index and then you don't have to get it again from the global state.

JennieDalgren (approved):
Good job with this project!

Let me know if you understand the index vs id thing or not.
I know you've been discussing this in you team?
The project is super nice styled and you did a good job with your checkboxes!
Maybe fix the little issue that you right now can add an empty todo, if you have time.

Keep up the good work 💪

PR #51, opened 2020-04-22
Todo project -ready for review plz

Tuliany:
Link to my deployed project:
https://todo-redux-by-tuliany.netlify.app/

It was kind of difficult in the beginning to get a hang of reducers, but after that, it was very fun. If I had more time, I would put an effort into making the stylecomponents work for the classNames., but I solved it by using CSS this time.

props.


code/src/App.js
```diff
@@ -1,9 +1,46 @@
...
+          <TaskSummary />
+          </section>
+            <Divider light />
+      </header>
+    
+      <section className="tasklist">
+      <TaskList />
+      </section>
+    </main>
+    </Provider>
```
idlefingers: Keep an eye on your indentation. It should be one new level of indentation for each new tag (or scope). Like this:
```jsx
export const App = () => {
  const day = moment().format('HH:mm | D/M')
  return (
    <Provider store={store}>
      <main>
        <header>
          <h1>TODO</h1>
          <section className="daysummary">
            <p>{day}</p>
            <TaskSummary />
          </section>
          <Divider light />
        </header>
    
        <section className="tasklist">
          <TaskList />
        </section>
      </main>
    </Provider>
  )
}
```

code/src/Components/TaskInput.js
```diff
@@ -0,0 +1,43 @@
...
+      className="task-input-text"
+      ></input>
+      <input
+      type="submit"
+      className="button"
+      value="Add task"
+      ></input>
+     </form>
+  
+  )
```
idlefingers: Indentation!

code/src/Components/TaskItem.js
```diff
@@ -0,0 +1,75 @@
...
+  const handleOnChange = (e) => {
+    dispatch(
+      todos.actions.setDone({
+      itemIndex: itemIndex,
+      done: !item.done,
+    })
+   )
+  }
+
+  const Wrapper = styled.div`
```
idlefingers, line 32: Styles should be defined outside your component, beneath the imports

code/src/Components/TaskItem.js
```diff
@@ -0,0 +1,75 @@
...
+      type="checkbox"
+      onChange={handleOnChange}
+      className="task-item-check"
+      checked={item.done ? "checked" : ""}>
+    </input>
+   
+
+    <span className="task-item-description">{item.description}</span>
+    <Button
+      className="my__unique__button"
```
idlefingers, line 67: You can skip setting class names and do this within styled components, by referring to another styled component (for this, `Button` needs to be defined before `Wrapper`), then in `Wrapper`:

```
const Wrapper = styled.div`
  // ... your normal styles

  &:hover ${Button} {
    transform: scale(1.8);
  }
`
```

Read more: https://styled-components.com/docs/advanced#referring-to-other-components

code/src/Components/TaskList.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const TaskList = () => {
+const list = useSelector((store) => store.todos.list)
+  
+  return (
+    <section>
+      <TaskInput />
+      {list.items.map((item, index) => {
+       return <TaskItem 
+       key="taskNum"
```
idlefingers, line 16: Keys should be unique in React. The `item`'s id would be a perfect thing to use for the key here - `key={item.id}`

More info: https://reactjs.org/docs/lists-and-keys.html

code/src/Components/TaskItem.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react'
+import { useDispatch, useSelector } from 'react-redux'
+
+import { todos } from '../Reducers/todos.js'
+import styled from 'styled-components'
+
+
+export const TaskItem = ({ itemIndex }) => {
+  
+  const item = useSelector((store) => store.todos.list.items[itemIndex])
```
idlefingers, line 10: Since you're passing in the `item` as a prop to this component, you needn't select it from redux again. You could just pick up the item prop in your props destructuring up in the component definition:

```
export const TaskItem = ({ item }) => {
```

idlefingers (approved):
Great work, and your app looks super nice! Well done! 👏 

For the next project, try to pay closer attention to your indentation and formatting of your code. Try to make it as clean and readable as possible.

Regarding props, they can be tricky to get the hang of! Think of it as passing arguments to a function. So if I had a component called `Fruit` and I want to tell that component what the name of the fruit is and how much it weighs, I might write this in JSX:

```
<Fruit name="Apple" weight="150g" />
```

And then when I create the fruit component, I can pick up the props object and access the attributes:

```
const Fruit = (props) => {
  // props.name = 'Apple'
  // props.weight = '150g'
}
```

So each attribute in JSX (`name="Apple"`, `weight="150g"`) becomes a key and value in the props object in your component: `{ name: "Apple", weight: "150g" }`.

Mix this together with [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) and you can pluck out individual props as variables in your component. We could re-write the `Fruit` component to this:

```
const Fruit = ({ name, weight }) => {
  // name = 'Apple'
  // weight = '150g'
}
```

Hopefully that helps to clear things up a bit! You can read up more on props here: https://reactjs.org/docs/components-and-props.html

PR #53, opened 2020-04-22
Review wanted - todo Frida

fridamaria:
Link to my deployed project: https://fridamaria-todos.netlify.app/

My reflections on how this week's project turned out: it was a fun project and I liked the way it turned out. Good to be able to follow Van's code along step by step as it helped me understand Redux better. This week's videos made it clearer as well, was very confused after the meet up. 

Things I'd like to have clarified or explained in more detail: I think I just need to look at the videos again and do some more reading. Redux is not as intuitive as React at this point, but I will get there :)


code/public/index.html
```diff
@@ -1,34 +1,22 @@
...
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <link rel="icon" href="%PUBLIC_URL%/2705.svg">
+  <link href="https://fonts.googleapis.com/css2?family=Playfair+Display:wght@900&display=swap" rel="stylesheet">
+  <meta property="og:url" content="https://fridamaria-todos.netlify.app/" />
+  <meta property="og:type" content="website" />
+  <meta property="og:title" content="Todo App" />
+  <meta property="og:description" content="Small todo app" />
+  <meta property="og:image" content="%PUBLIC_URL%/project-todo.png" />
```
idlefingers, line 13: 😎

code/src/components/EmptyState.js
```diff
@@ -0,0 +1,64 @@
...
+  margin: 2px 0;
+  font-size: 12px;
+  color: #3e3e2d;
+
+  @media (min-width: 768px) {
+    font-size: 14px;
+  }
+`
+
+export const EmptyState = () => {
```
idlefingers, line 50: Nice with something to show when the list is empty! 💯

code/src/components/Filters.js
```diff
@@ -0,0 +1,55 @@
...
+  const dispatch = useDispatch()
+
+  //Functions filtering active, done and all todos
+  const showActive = () => dispatch(todos.actions.showActive())
+  const showDone = () => dispatch(todos.actions.showDone())
+  const showAll = () => dispatch(todos.actions.showAll())
+
+  return (
+    <>
+      {itemsLength > 0 &&
```
idlefingers: Nitpicking here, but I think that if the component shouldn't render anything in a certain condition, then it's clearer to bail out before you start writing jsx. Something like this:

```jsx
export const Filter = () => {
  // ... all the initial stuff here

  if (itemsLength === 0) {
    return null
  }

  return (
    <FilterWrapper>
      ...
    </FilterWrapper>
  )
}
```

code/src/components/Filters.js
```diff
@@ -0,0 +1,55 @@
...
+        <FilterWrapper>
+
+          <Button
+            aria-label='Show active tasks'
+            onClick={showActive}
+            width={'100%'}
+            margin={'5px auto'}
```
idlefingers: There's no need for the curly braces when you're returning a string like this. You can just do `width="100%" margin="5px auto"` etc

code/src/components/TodoInput.js
```diff
@@ -0,0 +1,81 @@
...
+    <TodoForm onSubmit={handleOnSubmit}>
+
+      <TodoInputField
+        aria-label="Enter task here"
+        type='text'
+        onChange={e => setInputValue(e.target.value)}
+        value={inputValue}
+        placeholder='Add task...'
+        maxLength={50}
+      ></TodoInputField>
```
idlefingers: When a component (or div or any other element) has no children, you can write it as a self-closing tag:

```jsx
<TodoInputField
  aria-label="Enter task here"
  type='text'
  onChange={e => setInputValue(e.target.value)}
  value={inputValue}
  placeholder='Add task...'
  maxLength={50} />
```

code/src/components/TodoItem.js
```diff
@@ -0,0 +1,120 @@
...
+
+  //Removes todo from array
+  const onRemoveClicked = () => dispatch(todos.actions.removeTodo({ itemIndex }))
+
+  //Marks todo as done
+  const onCheckClick = () => dispatch(todos.actions.setDone({ itemIndex }))
+
+  return (
+    <>
+      {item.show &&
```
idlefingers: Same here with the `return null` trick

code/src/pages/Home.js
```diff
@@ -0,0 +1,31 @@
...
+import { todos } from '../reducers/todos'
+import { TodoList } from '../components/TodoList'
+
+// Variable combining reducers from todos
+const reducer = combineReducers({ todos: todos.reducer })
+
+// Persistence: Retrieves the existing state from localstorage if it exists
+const persistedState = localStorage.getItem('reduxState')
+  ? JSON.parse(localStorage.getItem('reduxState'))
+  : {};
```
idlefingers, line 13: Nice!

idlefingers (approved):
Amazing work Frida! Super nice & clean code, makes it a pleasure to read through. Looks like you have a great grasp of both react and redux to me. Awesome! ⭐

code/src/components/Filters.js
fridamaria: Ah, sounds logical! Will try it out :)

code/src/components/Filters.js
fridamaria: Makes sense, I'll remove it!

code/src/components/TodoInput.js
fridamaria: Ah, I missed that one :)

code/src/components/TodoItem.js
fridamaria: Done :)

PR #62, opened 2020-04-24
Project Todo - Mathias T - Review wanted

Nesiht:
Link to my deployed project:
[Link](https://flamboyant-swartz-55a72b.netlify.app/)

Fun project, spent way more hours in this one then in any other project. Focused a lot on functions, state and different ways to use props in styled components. If I had more time I would make it responsive and not mix vw/vh and pixel units when styling.


code/src/components/AddItem.js
```diff
@@ -0,0 +1,96 @@
...
+          category: category,
+          startDate: date.toLocaleDateString( 'se-SE', { dateStyle: 'short' }),
+          dueDate: 2021
+        }
+      })
+    )
+
+    setInputValue('')
+  }
+
```
JennieDalgren, line 73: Nice!
Two things here, you could set all the fields that are not selected by the user (such as id, startDate) in the reducer instead of here. Just passing the user inputs from this component. But this is totally fine!
The other thing is that you right now can add an empty todo. This you could easily check either by adding this line to the top of the handleSubmit function:
`if (!inputValue) return` Which will return from the function (stop it) if there is no input value. 
You could also make the input value a required field in the form or disable the submit button until something is put into the input.

code/src/components/DeleteItem.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { useDispatch } from "react-redux"
+import styled from 'styled-components'
+import { todo } from "../reducers/todo"
+
+const A = styled.a`
```
JennieDalgren, line 6: I might have called this `Link` instead of just an `A`... But no biggie!

code/src/components/DeleteItem.js
```diff
@@ -0,0 +1,26 @@
...
+      todo.actions.removeTodo({
+        itemIndex,
+      })
+    )
+  }
+
+  return (
+    <section>
+      <A onClick={onRemoveClicked}>🧨</A>
+    </section>
```
JennieDalgren, line 24: you don't need to wrap this in a section since the A is just one single element.

code/src/components/ListItem.js
```diff
@@ -0,0 +1,123 @@
...
+  border-radius: 8px;
+`
+const TodoItem = styled(Container)`
+  justify-content: space-evenly;
+  background-color: ${props => props.theme.main};
+  width: 25vw;
+  margin: ${props => props.backside ? "0px" : "8px"};
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
+  min-height: 111px;
+  background-color: ${props => props.bg == 'work' ? "#FFBB00" : props.bg == 'personal' ? '#7CBB00' : props.bg == 'other' ? '#00A1F1' : ''};
```
JennieDalgren, line 23: DON'T use only two `==` ALWAYS use `===` in javascript

code/src/components/ListItem.js
```diff
@@ -0,0 +1,123 @@
...
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
+  border-radius: 8px;
+`
+const TodoItem = styled(Container)`
+  justify-content: space-evenly;
+  background-color: ${props => props.theme.main};
```
JennieDalgren, line 18: nice work with dealing with props in styled components!

code/src/components/ListItem.js
```diff
@@ -0,0 +1,123 @@
+import React from 'react'
+import { useSelector } from 'react-redux'
+import Flippy, { FrontSide, BackSide } from 'react-flippy'
```
JennieDalgren, line 3: flippy! nice thing

code/src/components/ListItem.js
```diff
@@ -0,0 +1,123 @@
...
+
+              <TodoSection>
+                <h1>
+                  {item.description}
+                </h1>
+              </TodoSection>
+
+              <InfoSection>
+                <span></span>
+              </InfoSection>
```
JennieDalgren, line 86: What is this info section used for?

code/src/components/SetDone.js
```diff
@@ -0,0 +1,32 @@
...
+import { useDispatch } from "react-redux"
+import styled from 'styled-components'
+import { todo } from "../reducers/todo"
+
+const A = styled.a`
+  cursor: pointer;
+`
+
+export const SetDone = ({itemIndex}) => {
+  const [updateDone, SetupdateDone] = useState(true)
```
JennieDalgren, line 11: the method here should be camelCased: `setUpdateDone`

code/src/components/SetDone.js
```diff
@@ -0,0 +1,32 @@
...
+  const dispatch = useDispatch()
+
+  const onDoneClicked = (e) => {
+    SetupdateDone(!updateDone)
```
JennieDalgren, line 15: again, camelCased: `setUpdateDone`

JennieDalgren: no need to have the `e` as a param in this function, right?

code/src/components/SetDone.js
```diff
@@ -0,0 +1,32 @@
...
+      })
+    )
+  }
+
+  return (
+    <section>
+      <A onClick={onDoneClicked}>
+        {updateDone ? "✔️" : "❌"}
+      </A>
+    </section>
```
JennieDalgren, line 30: no need for the section wrapper here, the A is enough

code/src/components/SettingsContainer.js
```diff
@@ -0,0 +1,21 @@
...
+  justify-content: space-evenly;
+  align-items: flex-start;
+  background-color: DodgerBlue;
+`
+
+export const SettingsContainer = () => {
+  return (
+    <Main>
+    <ShowSummery />
+    <ShowFilter />
```
JennieDalgren, line 18: indentation 😎

code/src/components/ShowFilter.js
```diff
@@ -0,0 +1,45 @@
...
+  margin-bottom: 10px;
+`
+const A = styled.a`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 30px;
+  height: 30px;
+  border-radius: 100%;
+  background-color: ${props => props.default ? 'white' : props.work ? '#FFBB00' : props.personal ? '#7CBB00' : props.other ? '#00A1F1' : 'white'};
```
JennieDalgren, line 27: nice work with the ternary, although it can be a bit hard to read when there are multiple ternaries nested like this.

code/src/components/TodoContainer.js
```diff
@@ -0,0 +1,26 @@
...
+export const TodoContainer = () => {
+  return(
+    <>
+      <SettingsContainer />
+      <Container>
+        <AddItem />
+        <ListItem />
+      </Container>
+    </>
+  )
```
JennieDalgren, line 25: this could be written as an implicitly return:
```
export const TodoContainer = () => (
    <>
      <SettingsContainer />
      <Container>
        <AddItem />
        <ListItem />
      </Container>
    </>
  )
```

code/src/components/ToggleItem.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+export const ToggleItem = () => {
+  return (
+    <p>ToggleItem!</p>
+  )
```
JennieDalgren, line 6: same here:
```
export const ToggleItem = () => <p>ToggleItem!</p>
```

code/src/reducers/todo.js
```diff
@@ -0,0 +1,91 @@
...
+      description: 'Social distancing',
+      category: 'other',
+      done: false,
+      startDate: '2020-04-15',
+      dueDate: '2020-04-30'
+    }
+    ]
+  }
+}
+
```
JennieDalgren, line 51: ⭐ nice and great comments

JennieDalgren (approved):
Really good job with this project! ⭐️

You've really managed to get a good understanding on how to deal with both React and Redux. Great job splitting into smaller components! Also, great job using styled components.

Keep up the good work 💪🏼

PR #65, opened 2020-04-24
Todo app Jihyun An 

jihyeona:
Link to my deployed project:
https://practical-bassi-d5a5e9.netlify.app/

I'm happy with how it turned out, but it feels like there's a bit room there left to push the project even further. I'm happy that I got to train using styled components more. It was also a good exercise to understand more how the Redux works and where to pass in the props.  

Maybe how to sort the tasks list in relation to how much time is left until their due date? I got too tired in my head to figure out the logic for that one.

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -17,7 +17,10 @@ if (persistedStateJSON) {
 }
 console.log(`persistedState: ${persistedState}`)
 
-const store = configureStore({ reducer, persistedState })
+const store = configureStore({
+  reducer,
+  preloadedState: persistedState
+})
 
 store.subscribe(() => {
   localStorage.setItem('reduxState', JSON.stringify(store.getState()))
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,23 @@
 import React from 'react'
+import { Provider } from 'react-redux'
+import { combineReducers, configureStore } from '@reduxjs/toolkit'
+import { notepad } from 'reducers/notepad'
+import { TaskSummary } from 'components/TaskSummary'
+import { AddTaskForm } from 'components/AddTaskForm'
+import { TaskList } from 'components/TaskList'
+
+const reducer = combineReducers({
+  notepad: notepad.reducer
+})
+
+const store = configureStore({ reducer })
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <Provider store={store}>
+      <TaskSummary />
+      <AddTaskForm />
+      <TaskList />
+    </Provider>
   )
 }
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,16 +1,28 @@
 import React from 'react'
 import { Provider } from 'react-redux'
-import { combineReducers, configureStore } from '@reduxjs/toolkit'
+import { combineReducers, configureStore, createStore } from '@reduxjs/toolkit'
 import { notepad } from 'reducers/notepad'
 import { TaskSummary } from 'components/TaskSummary'
 import { AddTaskForm } from 'components/AddTaskForm'
 import { TaskList } from 'components/TaskList'
+import { CustomCheckBox } from 'components/CustomCheckBox'
 
 const reducer = combineReducers({
   notepad: notepad.reducer
 })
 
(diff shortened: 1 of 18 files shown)


code/src/App.js
```diff
@@ -1,9 +1,34 @@
...
+if (persistedStateJSON) {
+  persistedState = JSON.parse(persistedStateJSON)
+}
+console.log(`persistedState: ${persistedState}`)
+
+const store = configureStore({ reducer, persistedState })
+
+store.subscribe(() => {
+  localStorage.setItem('reduxState', JSON.stringify(store.getState()))
+})
```
HIPPIEKICK: Nice that you're trying out storing the todos! I read a little more about `configureStore`, and apparently it doesn't have `persistedState` as a config option. However, it does have something called `preloadedState` which we can use. To get this to work, just define your store like this:
```
  reducer,
  preloadedState: persistedState
})
```
If you want to read more about configureStore's options, you can take a look here: [https://redux-toolkit.js.org/api/configureStore#preloadedstate](https://redux-toolkit.js.org/api/configureStore#preloadedstate)

code/src/components/AddTaskForm.js
```diff
@@ -0,0 +1,89 @@
...
+
+  const handleSubmit = (event) => {
+    event.preventDefault()
+    dispatch(notepad.actions.addTask({ name, description, dueDate, category, done }))
+    setName('')
+    setDescription('')
+    setDueDate(new Date())
+    setCategory('')
+    setDone(false)
+  }
```
HIPPIEKICK, line 30: Right now you can submit empty todos, maybe check the input before the dispatch (if the input is empty or not) to prevent this.

code/src/components/AddTaskForm.js
```diff
@@ -0,0 +1,89 @@
...
+                value={description}
+                onChange={(event) => setDescription(event.target.value)}
+              />
+            </label>
+            <label>
+              Due Date:
+            <DatePicker
+                selected={dueDate}
+                onChange={(date) => setDueDate(date)}
+              />
```
HIPPIEKICK, line 67: ⭐

code/src/components/AddTaskForm.js
```diff
@@ -0,0 +1,89 @@
...
+                value={category}
+                onChange={(event) => setCategory(event.target.value)}
+              >
+                <option value='option'>option</option>
+                <option value='home'>home</option>
+                <option value='work'>work</option>
+                <option value='fun'>fun</option>
+                <option value='coding'>coding</option>
+                <option value='training'>training</option>
+              </Category>
```
HIPPIEKICK, line 81: Nice!

code/src/components/Task.js
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <>
+      <ListContainer>
+        <label>
+          <CustomCheckBox
+            isChecked={props.item.done}
+            onChangeHandler={handleCheckBox} />
+        </label>
+        <span onClick={showDetailHandler}>{props.item.name} by {moment(props.item.dueDate).format('dddd, MMM Do')}</span>
```
HIPPIEKICK, line 38: Nice that you're using moment!

HIPPIEKICK (approved):
Awesome that you're challenging yourself with the stretch goals - so much functionality! Also nice to see that you're using some external libraries (moment, datepicker and such) and are able to implement them well. One thing to think about in the future, is to write a bit more descriptive commit messages.

Regarding sorting the items - since Redux want's you to only mutate the state through the reducer, you have to make a copy of the array before you can sort it somewhere else. Since you just want to change how it's shown in the app, you can do it like this:
```
const tasks = useSelector((state) => state.notepad.items)
const tasksCopy = tasks.slice()
const sortedTasks = tasksCopy.sort((a, b) => a.dueDate - b.dueDate)
```
and then map over it in TaskList.js. Here you can read more about it the issue if you like: [https://stackoverflow.com/questions/53420055/error-while-sorting-array-of-objects-cannot-assign-to-read-only-property-2-of/53420326](https://stackoverflow.com/questions/53420055/error-while-sorting-array-of-objects-cannot-assign-to-read-only-property-2-of/53420326)

Good job Jihyun! ⭐

PR #74, opened 2020-06-02
Todo app Åsa Andersson

aasa72:
Link to my deployed project: https://romantic-brahmagupta-cdc2f1.netlify.app/


HIPPIEKICK (approved):
Well done! Just a little FYI, since the index of an item can change, it's better to use some sort of ID when dealing with the done state. Just something to think about in the future :)

PR #77, opened 2020-06-10
Todo Project Viktoria Berglund

VickBe:
Link to my deployed project:https://practical-torvalds-64a6e7.netlify.app/


HIPPIEKICK (changes requested):
Nice with local storage! Overall, it looks very good but you missed one requirement:
- Your app should show a count of either all todos, or all uncomplete todos

Also, there's a little bug appearing when you check the first todo as complete, and then remove it - the second todo gets checked instead. Since the index of an item can change, it's better to use some sort of ID when dealing with the complete state. Just something to think about in the future :)

HIPPIEKICK (approved):

PR #83, opened 2020-11-25
Project ToDos - Vanessa Sue Smith

VanessaSue27:
Link to my deployed project: https://vane-todos.netlify.app/

Steps on how I built the app are listed on the ReadMe and also through comments in the code itself 😁

I have been in the lookout for a good To Do App since I work with daily to do lists: so this was a great opportunity to build an app tailored to my own needs 😋.
I had in mind from the beginning that I wanted to sort the tasks by category since that is how I usually write down my daily to dos. So the 4 main categories are the ones I'm practically using in my daily life.

It was challenging deciding where to start since this is the first time we create the Store from scratch, so I decided to follow the hints given in the project brief and then I could take over from there.

I tried to create the unique IDs for the tasks with the Math.max() method Maks showed, but this created a bug when deleting all tasls, since it would give id: null - so had to change it back to create the ID by using Date.now()

I wanted to challenge myself and style this app only using Styled components: this presented two challenges: customizing the checkboxes and the select dropdown.
It was a bigger challenge getting the custom checkboxes to work since they not only include style but also functionality. It ended up being a lot of code, but a good learning experience and reference for the future.

It was super fun being able to implement vanilla JS in the reducers, that usually is my favorite part, but have to admit I also enjoyed creating my styled-components library. It ended up being a fun alternative way of styling.

// Vanessa 🍉


code/src/components/AddToDoInput.js
```diff
@@ -0,0 +1,68 @@
+import React, { useState } from 'react';
+import { useDispatch } from 'react-redux';
```
destinyleia, line 2: I think the "import { useDispatch } from 'react-redux';" and the "import { todolist } from 'reducers/todolist';" dependencies should have a blank space between each other. I think it works as with components and that reducers/todolist is a local dependency. 📱

Please correct me if I'm wrong! 😁

destinyleia (approved):
OMG VANESSA! You are too good to be true! ⭐️

I love your idea with different topics for different todos and how you always explain your thinking with comments! ❤️

I saw one thing that I know is a careless error that I commented but I don't se any other suggestions to change your code... I give it to the code coaches! 😅

Your code is magical! It was hard reviewing this...😁

**🔵  Blue Level (Minimum Requirements)**

- [X] Your app should list all todo tasks - completed or uncompleted

- [X] You should be able to mark an uncompleted task as complete (and change it back to uncomplete)

- [X] You should be able to add and remove tasks

- [X] Your app should show a count of either all todos, or all uncomplete todos (or both)

I will approve this projects because I think it fulfills all the blue requirements! 😄🥳
Great job as always code queen Vanessa! ❤️

code/src/App.js
```diff
@@ -1,9 +1,67 @@
...
+    // are called within this provider will have access to the data
+    // in the store
+    <Provider store={store}>
+      <Header />
+      <AddToDoInput />
+      
+      <StudiesTasks />
+      <HouseChoreTasks />
+      <ShoppingList />
+      <OtherTasks />
```
HIPPIEKICK, line 60: Would be nice if these were conditionally rendered, i.e. if there are only `StudiesTasks`, don't show the other headings

code/src/components/AddToDoInput.js
```diff
@@ -0,0 +1,68 @@
...
+  // in data to the global state regargind the new task to add to the list. So I pass in
+  // an object containig the data on the new task to be added, this is the payload
+  const handleOnSubmit = (event) => {
+    event.preventDefault();
+    
+    // Setting the id for the new tasks with a timestamp, this will create a quite
+    // unique ID and chances of it repeating are super low.
+    // the only tasks with a set id are the ones included in the initial state
+    dispatch(todolist.actions.addToDo({
+      id: Date.now(),
```
HIPPIEKICK, line 28: Would be nice with a time stamp as well :)

code/src/components/AddToDoInput.js
```diff
@@ -0,0 +1,68 @@
...
+  // all items in the to do list
+  return (
+    <AddToDoForm onSubmit={handleOnSubmit}>
+      <LabelText>
+        Add To Do: 
+        <AddToDoInputField 
+          type="text"
+          onChange={event => setNewTodoItem(event.target.value)}
+          value={newTodoItem}
+          required >
```
HIPPIEKICK, line 52: Boolean attributes should be written before other attributes

code/src/components/Header.js
```diff
@@ -0,0 +1,33 @@
...
+  const toDoList = useSelector((state) => state.todolist.items);
+  const missing = toDoList.filter(x => x.complete === false);
+  const missingCount = missing.length;
+
+  const months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sept', 'Oct', 'Nov', 'Dec'];
+  const date = new Date();
+  const mm = date.getMonth();
+  const month = months[mm];
+  const dd = date.getDate();
+  const specificWeekday = date.toLocaleDateString('en-US', {weekday: 'long'});
```
HIPPIEKICK, line 19: Have you checked in [moment js](https://momentjs.com/)?

code/src/styled-components/AddToDoComponents.js
```diff
@@ -0,0 +1,54 @@
+import styled from 'styled-components';
```
HIPPIEKICK, line 1: Nice with styled components ⭐

PR #114, opened 2020-11-29
Project-todos Emelie Svensson

emeliesv:
Link to my deployed project:
https://project-todos-emelie.netlify.app/

I'm satisfied with the styling, even though I probably could work on it a bit more. But most importantly: I'm really glad that I understand Redux and feel like I grasp what it does, how I can use it and how powerful it actually is! So overall, I'm happy with this!


code/public/index.html
```diff
@@ -1,34 +1,27 @@
...
+  <meta property="og:image" content="./Todos.jpg">
+  <meta property="og:url" content="https://project-todos-emelie.netlify.app/">
+  <meta name="twitter:title" content="The very important Todo site">
+  <meta name="twitter:description" content="This is a site where you can add all your important todos!">
+  <meta name="twitter:image" content="./Todos.jpg">
+  <meta name="keywords" content="Todo, tasks, lists, todoapp, react, javascript, styled components,">
+  <meta name="robots" content="index, follow">
+  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+  <meta name="language" content="English">
+  <meta name="author" content="Emelie Svensson">
```
thereselinden: Nice with the meta tags 💯 Maybe add a favicon for future projects? 😀

code/src/App.js
```diff
@@ -1,9 +1,30 @@
...
+let persistedState = {}
+if (persistedStateJSON) {
+  persistedState=JSON.parse(persistedStateJSON)
+}
+
+const store = createStore(reducer, persistedState, window.__REDUX_DEVTOOLS_EXTENSION__&& window.__REDUX_DEVTOOLS_EXTENSION__())
+
+store.subscribe(() => {
+  localStorage.setItem('taskState', JSON.stringify(store.getState()))
+})
```
thereselinden: Great job with the localStorage!! ⭐️

code/src/App.js
```diff
@@ -1,9 +1,30 @@
...
+const reducer = combineReducers({tasks: tasks.reducer,})
+//Here we'll use localstorage to save the current state
+
+const persistedStateJSON = localStorage.getItem('taskState')
+let persistedState = {}
+if (persistedStateJSON) {
+  persistedState=JSON.parse(persistedStateJSON)
+}
+
+const store = createStore(reducer, persistedState, window.__REDUX_DEVTOOLS_EXTENSION__&& window.__REDUX_DEVTOOLS_EXTENSION__())
```
thereselinden: Don't think you need to use `window.__REDUX_DEVTOOLS_EXTENSION__&& window.__REDUX_DEVTOOLS_EXTENSION__()`

The store could be written like this 
`const store = configureStore({
  reducer,
  preloadedState: persistedState,
});`

code/src/Components/Checkbox.js
```diff
@@ -0,0 +1,39 @@
...
+
+import {
+  HiddenCheckbox,
+  StyledCheckbox,
+  CheckboxContainer,
+  Icon,
+} from '../Styles/CostumCheckbox'
+
+export const Checkbox = ({ done, onChangeHandler }) => {
+  const [checked, setChecked] = useState(done)
```
thereselinden, line 11: Instead of creating local state, you could use the completion status within the global state (reducers). 
In your dispatch when adding a task you send information about "done" which start off as false.  You could then  call your `toggleDone ` action in a dispatch from this component. Or the JSX (line 19-26) could be within your `customCheckbox` located in styles folder. Then that component will be even more reusable.

code/src/Components/Header.js
```diff
@@ -0,0 +1,21 @@
...
+export const Header = () => {
+  const items = useSelector((store) => store.tasks.items)
+  const completedTasks = items.filter((item) => item.done)
+
+  return (
+    <>
+      <Heading>Todo's</Heading>
+      {completedTasks.length<items.length ? <SubHeading>You have completed {completedTasks.length} out of {items.length} tasks</SubHeading>
+      : <Wrapper><SubHeading>You have done all your tasks! Grab an icecream and celebrate!</SubHeading>
+      <Image src={'./lottieanimation.gif'} alt='animation of Icecream'/></Wrapper>}
```
thereselinden: For better readability, I think line breaks and indentation would be great here. 😀

code/src/library/Containers.js
```diff
@@ -0,0 +1,55 @@
...
+display: grid;
+grid-template-columns: 1fr;
+grid-template-rows: min-content; 
+border-radius: 10px;
+`}
+justify-content: center;
+width: 100%;
+margin: 15px;
+max-width: 90%;
+}
```
thereselinden: Indentation would be great here for better readability.

code/public/index.html
```diff
@@ -1,34 +1,27 @@
...
-    <!--
-      Notice the use of %PUBLIC_URL% in the tags above.
-      It will be replaced with the URL of the `public` folder during the build.
-      Only files inside the `public` folder can be referenced from the HTML.
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta property="og:title" content="The very important Todo site">
+  <meta property="og:description" content="This is a site where you can add all your important todos!">
+  <meta property="og:image" content="./Todos.jpg">
```
thereselinden: I don't think this path is correct, the image is not showing.

code/src/App.js
```diff
@@ -1,9 +1,30 @@
...
+  localStorage.setItem('taskState', JSON.stringify(store.getState()))
+})
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <Provider store={store}>
+      <Summary/>
```
thereselinden: Since this is the only component mounting within the app component, maybe summary is not a very descriptive name of this component.

code/src/Components/Header.js
thereselinden: Nice with then animation 😃

code/src/Components/Item.js
```diff
@@ -0,0 +1,35 @@
...
+import { useDispatch } from 'react-redux'
+
+import { tasks } from '../Reducers/tasks'
+
+import { Button } from '../library/Button'
+import { Listelement } from '../Styles/Listelement'
+import { Checkbox } from './Checkbox'
+import { Text, Span } from '../Styles/Texts'
+
+import moment from 'moment'
```
thereselinden, line 11: I think convention is that everything imported from "outside" should be after each other. This could be moved to line 3.

code/src/Components/Item.js
```diff
@@ -0,0 +1,35 @@
...
+
+export const Item = ({ item }) => {
+  const dispatch = useDispatch()
+  const [removeBtn] = useState(true)
```
thereselinden: I never seen this, what is the use of just having a variable in the useState hook? Is const a better solution since you are not changing its value on line 26? It's only for styling purposes? 
Maybe instead you can create different buttons within your lib. Example: 
```
export const RemoveButton
export const AddButton
export const ClearButton 
```
And reuse the styling that is the same. 

I seen that you are doing this in other components too.

code/src/Components/Item.js
```diff
@@ -0,0 +1,35 @@
...
+
+  const handleCheckboxClick = () => {
+    dispatch(tasks.actions.toggleDone({ id: item.id}))
+  }
+
+  return (
+    <Listelement>
+      <Checkbox done={item.done} onChangeHandler={handleCheckboxClick}/>
+      <Text stroke={item.done ? true : ''}>{item.text}<Span>Created: {moment(item.created).fromNow()}</Span></Text> {/* I get an error when the stroke is set to true, but don't know how to solve it */}
```
thereselinden, line 23: I don't see an error when I check the todo as done.

code/src/Components/List.js
```diff
@@ -0,0 +1,23 @@
...
+import { Item } from './Item';
+import { ListWrapper } from '../library/Containers';
+
+export const List = () => {
+  //This useSelector helps us get a hold of the items inside our store (tasks)
+
+  const items = useSelector((store) => store.tasks.items);
+
+  return (
+    <>
```
thereselinden: You don't need this fraction since the `ListWrapper` is wrapping the return content.

code/src/Components/UserInput.js
```diff
@@ -0,0 +1,44 @@
...
+import { useDispatch } from 'react-redux'
+
+import { tasks } from '../Reducers/tasks'
+
+import { Button } from '../library/Button'
+import { UserInputBox, InputText } from '../library/Containers'
+
+export const UserInput = () => {
+  const [task, setTask] = useState('')
+  const [addBtn] = useState(true)
```
thereselinden: See comment above regarding this.

code/src/Components/UserInput.js
```diff
@@ -0,0 +1,44 @@
...
+
+  const dispatch = useDispatch()
+
+  const handleSubmit = (event) => {
+    event.preventDefault()
+    dispatch(
+      tasks.actions.addItem({
+        taskInfo: {
+          id: Date.now(),
```
thereselinden, line 20: A more used way of handling id is to use uuid library. You will then have to do a npm install and import that library to this component.  
Installation: `npm install uuid` I
mport: `import { v4 as uuidv4 } from 'uuid';` 
Call function:  `uuidv4() `  instead of Date.now() on line 21.

code/src/Pages/Summary.js
```diff
@@ -0,0 +1,40 @@
...
+import { List } from '../Components/List'
+import { Header } from '../Components/Header'
+import { UserInput } from '../Components/UserInput'
+import { Button } from '../library/Button'
+import { Main, Wrapper } from '../library/Containers'
+
+export const Summary = () => {
+  const [clearBtn] = useState(true)
+  const [outerWrapper] = useState(true)
+  const [noList] = useState(true)
```
thereselinden: See comment regarding this above.

code/src/Reducers/tasks.js
```diff
@@ -0,0 +1,30 @@
...
+      state.items = state.items.filter(task=>task.id!==id)
+    },
+    removeAll: (state) => {
+      state.items = []
+    },
+    toggleDone: (state, action) => {
+      const { id } = action.payload
+      const foundItem = state.items.find(item => item.id === id)
+      if (foundItem) {
+        foundItem.done = !foundItem.done
```
thereselinden, line 26: 💯

code/src/Styles/CostumCheckbox.js
```diff
@@ -0,0 +1,39 @@
...
+border-radius: 3px;
+transition: all 150ms;
+${Icon} {
+    visibility: ${({checked}) => checked ? 'visible' : 'hidden'}
+}
+`
+
+export const CheckboxContainer = styled.div`
+display: inline-block;
+vertical-align: middle;`
```
thereselinden: Indentation and line break would be great here for better readability.

code/src/Styles/Listelement.js
```diff
@@ -0,0 +1,11 @@
...
+
+export const Listelement = styled.li`
+list-style: none;
+background: #B8CEBF;
+width: 100%;
+display: grid;
+grid-template-columns: repeat(4, 1fr);
+padding: 2px;
+margin: auto;
+`
```
thereselinden: Indentation is missing here.

code/src/library/Button.js
```diff
@@ -0,0 +1,32 @@
...
+    clearBtn &&
+    `
+    background: #34442F;
+    color: #CEE3D5;
+    border-radius: 10px;
+    width: 70%;
+    font-weight: 600;
+    margin: 10px;`}
+    border: none;
+`
```
thereselinden: See comment above regarding a suggestion on how to split this button.

code/src/library/Containers.js
```diff
@@ -0,0 +1,55 @@
+import styled from 'styled-components'
+
+export const Main = styled.main`
+${({noList}) => noList?`height: 100vh;`:`height:100%;`}
```
thereselinden: I can't see that this work. Height is 100vh even though I have added todos to the list.

code/src/library/Containers.js
```diff
@@ -0,0 +1,55 @@
...
+background-image: linear-gradient(45deg, #4e6048, #70836d, #93a895, #b8cebf);
+display: grid;
+grid-template-column: 1fr;
```
thereselinden: No need to set grid-template-column, since this will be per default depending on the structure of the code below.

code/src/library/Containers.js
```diff
@@ -0,0 +1,55 @@
...
+grid-template-rows: min-content;
+justify-content: center;
+padding: 10px;
+`
+
+export const Wrapper = styled.section`
```
thereselinden, line 12: Here you are using both flex and grid in the same styled component. You are setting width to 100% and max-width to 90%. 
Depending if  I have todos in my list or no todos in the list the wrappers size differ. If I add a lot of todos due to height of 100vh in main the list will not fit the main wrapper.

thereselinden (changes requested):
Hey Emelie! Nice work this week, I really liked your animation 😃🍦
I can see that you put in a lot of work this week sending props and took advantage of styled components and are using LocalStorage 💯.

General feedback: 
- Not consistent using semicolons.
- Indentations and line breaks 
- No media queries used, the text is quite small on larger devices
- You are using useState but not changing its value - consider changing to variables instead. 
- Maybe add a disable functionality for the add todo button, right now I can add todos even though I have not typed anything. 
- When I add several todos the list does not fit the main wrapper due to height set to 100vh. 

Please choose whether to use semicolons or not and add indentation. 

Keep up the great work! 🥰

code/src/App.js
emeliesv: Very weird that it worked without the `window.__REDUX_DEVTOOLS_EXTENSION__&& window.__REDUX_DEVTOOLS_EXTENSION__()` but not when I first had to add it 🤔 Also, what is preloadedState in that case? 😊

code/src/App.js
emeliesv: 😊

code/src/Components/Checkbox.js
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from 'react'
+
+import {
+  HiddenCheckbox,
+  StyledCheckbox,
+  CheckboxContainer,
+  Icon,
+} from '../Styles/CostumCheckbox'
```
EmOlsz, line 8: No need to name styles folder with capital letter. `/style` is good enough :)

code/src/Components/Checkbox.js
```diff
@@ -0,0 +1,39 @@
...
+    </CheckboxContainer>
+  )
+
+  return (
+    <div>
+      <label>
+        <CustomizedCheckbox
+          checked={checked}
+          onChange={handleOnChange}
+        ></CustomizedCheckbox>
```
EmOlsz: It should rather be self-closing component

code/src/Components/List.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const List = () => {
+  //This useSelector helps us get a hold of the items inside our store (tasks)
+  const items = useSelector((store) => store.tasks.items)
+
+  return (
+      <ListWrapper>
+        {items.map((item) => {
+            return(
+              <Item key={item.id} item={item} />
```
EmOlsz, line 14: Or just `{items.map((item) => <Item key={item.id} item={item} />}`

code/src/Components/UserInput.js
```diff
@@ -0,0 +1,44 @@
...
+import { useDispatch } from 'react-redux'
+
+import { tasks } from '../Reducers/tasks'
+
+import { Button } from '../library/Button'
+import { UserInputBox, InputText } from '../library/Containers'
+
+export const UserInput = () => {
+  const [task, setTask] = useState('')
+  const addBtn = true
```
EmOlsz: Is there a reason for this variable?

code/src/Components/UserInput.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  return (
+      <UserInputBox onSubmit={handleSubmit}>
+        <InputText
+          rows={4}
+          placeholder='Add task'
+          onChange={(event) => setTask(event.target.value)}
+          value={task}
+        ></InputText>
```
EmOlsz, line 37: Input should rather be self-closing tag

code/src/Pages/Home.js
```diff
@@ -0,0 +1,33 @@
...
+import { tasks } from '../Reducers/tasks'
+
+import { List } from '../Components/List'
+import { Header } from '../Components/Header'
+import { UserInput } from '../Components/UserInput'
+import { Button } from '../library/Button'
+import { Main, Wrapper, OuterWrapper } from '../library/Containers'
+
+export const Home = () => {
+  const clearBtn = true
```
EmOlsz: I don't see a reason for this variable as well, you could just pass boolean value to component itself, no need to wrap it in variable if we don't overwrite it anywhere

PR #121, opened 2020-11-29
Todos - Sara Stjernberg

stjernberg:
Link to my deployed project:
https://saras-todo-list.netlify.app/

It went quite well, although a bit hard to understand everyting.  It was good to do all the resux stuff from scratch. 


code/src/App.js
```diff
@@ -1,9 +1,38 @@
...
+        <Header />
+        <AddTask />
+        <TaskList />
+        <RemoveTasks />
+      </Wrapper>
+    </Provider>
   )
 }
+
+const Wrapper = styled.main`
```
karinnordkvist, line 30: Nice that you've used semantic tags- other than div. 😊👍

code/src/App.js
```diff
@@ -1,9 +1,38 @@
...
+  background-image: url(${backgroundImg});
+  background-size: cover;
+  width:600px;
+  height:800px;
```
karinnordkvist: Maybe use scaling measurements for wrapping tags instead of fixed pixel-measurements. On mobile your background does not cover the entire width, nor is it centered. From a design-perspective- what is the intention? 😊 If you wish to cover the entire screen, try:
`width: 100vw;, 
height: 100vh;`

Skärmavbild 2020-11-30 kl  21 48 23

code/src/components/AddTask.js
```diff
@@ -0,0 +1,62 @@
+import React, { useState } from 'react'
+import styled from 'styled-components'
+import { useDispatch } from 'react-redux'
+
+import { tasks } from 'reducers/tasks'
+
+export const AddTask = () => {
+  const dispatch = useDispatch()
+  //State for input from textbox
+  const [value, setValue] = useState("")
```
karinnordkvist: Maybe name this state something more suggestive, like textInput or todo? 😊

code/src/components/AddTask.js
```diff
@@ -0,0 +1,62 @@
...
+  height: 30px;
+`
+const AddButton = styled.button`
+  background-color: #e6e6e6; 
+  padding: 5px;
+  border-radius: 50%;
+  height: 30px;
+  width: 30px;
+  align-items: center;
+  justify-content: center;
```
karinnordkvist: I think you need to add `display:flex;` to be able to make use of these. I tried adding it in the developer tools, and it did center the plus-sign within the button. 😊

code/src/components/CheckBox.js
```diff
@@ -0,0 +1,70 @@
...
+  const [checked, setChecked] = useState(isChecked)
+
+  const handleOnChange = event => {
+    onChangeHandler()
+    setChecked(event.target.checked)
+  }
+
+  const Checkbox = ({ className, checked, ...props }) => (
+    <CheckboxContainer>
+      <HiddenCheckbox checked={checked} {...props}></HiddenCheckbox>
```
karinnordkvist, line 15: Nice that you used the spread operator for the props. 👍

code/src/components/CheckBox.js
```diff
@@ -0,0 +1,70 @@
+import React, { useState } from 'react'
```
karinnordkvist, line 1: Love the custom checkbox. Looks great! 🤩

code/src/components/Header.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const HeaderWrapper = styled.section`
+  width:600px;
+  height: 100px;
+  background: #f1f1f1;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  align-items: center;
```
karinnordkvist, line 23: No need for 2 of these I think 😅

code/src/components/RemoveTasks.js
```diff
@@ -0,0 +1,54 @@
...
+import styled from 'styled-components'
+import { useDispatch, useSelector } from 'react-redux'
+
+import { tasks } from 'reducers/tasks'
+
+
+export const RemoveTasks = () => {
+
+  const items = useSelector((store) => store.tasks.items);
+  const enabled = items.length > 0;
```
karinnordkvist: Was a bit unsure what this was until I found the variable in the buttons. Smart way of setting the buttons to disabled. However, would have been nice with either a more descriptive name like enableButtons or similar, or a comment above 😊

code/src/components/RemoveTasks.js
```diff
@@ -0,0 +1,54 @@
...
+    </DeleteCheckedButton>
+    </ButtonWrapper>
+  );
+};
+
+
+const ButtonWrapper = styled.section`
+  width: 100%;
+  position: absolute;
+  align-items: center;
```
karinnordkvist: Two of these again 😅

code/src/components/Task.js
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+
+  return (
+    <TaskContainer>
+      <CheckBox
+        isChecked={item.complete}
+        onChangeHandler={handleCheckboxClick}
+      />
+      <TaskText line={item.complete ? 'line-through' : null}>{item.text}</TaskText>
```
karinnordkvist: Hmm, I might be tired, but I think that the CSS should be in the styling below, and this part rather be a variable? Just to make the code a bit easier to understand 😊

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,50 @@
...
+    initialState: {
+    items: [],
+     
+   
+  },
+ 
+    reducers: {
+      addTask:(state, action) =>{
+        const { text } = action.payload
+        state.items.push({
```
karinnordkvist: To avoid mutation of the initial state we should not use the push-method, I think. You could instead try to create a new variable array, spread the initial state and then add the new task. I found this article and thought it was pretty good at explaining it.. 😊 (https://medium.com/javascript-in-plain-english/how-to-add-to-an-array-in-react-state-3d08ddb2e1dc) and this one was really good, too (https://medium.com/@thejasonfile/using-the-spread-operator-in-react-setstate-c8a14fc51be1) 📕🤓

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,50 @@
...
+        const matchTask = state.items.find((item) => item.id === action.payload)
+      
+             if (matchTask) {
+              matchTask.complete = !matchTask.complete
+        };
+      },
+    }
+})
+
+ // setDone: (state, action) => {
```
karinnordkvist: Maybe remove this if it's not used? 😊

code/src/components/AddTask.js
```diff
@@ -0,0 +1,62 @@
...
+  //State for input from textbox
+  const [value, setValue] = useState("")
+
+
+  //Handle submit function to dispatch task
+  const handleSubmit = (event) => {
+    event.preventDefault()
+
+    //Dispatch the action to save the new task.
+    dispatch(
```
karinnordkvist, line 17: I noticed that there was a possibility of adding a task without any text in it. Maybe add some sort of conditional in here to prevent that empty tasks end up in the list. Like so:
Skärmavbild 2020-11-30 kl  22 49 07

karinnordkvist (changes requested):
Good job on this weeks project, Sara! You write nice code with a lot of space in it, which makes it nice and readable. 😊👍

The only thing I noticed was missing from the blue level was the counter. There were also a small bug in the code, allowing the user to add empty tasks 🙊. All in all, there are a few small things to fix, such as code written twice, some naming-ideas and maybe - if you have time - take a look at the immutability-articles to avoid future mutations? 

I also noticed that your styled components declarations were in the top on some components, and in the bottom on some- maybe chose where they belong and put all in the same place. Just nitpicking a little 😅

But again, great job Sara! 🎈

code/src/components/AddTask.js
HIPPIEKICK: Yup, this need to be fixed 👍

HIPPIEKICK: Also, the design goes a little whacky after submitting many todos 🙈 
Skärmavbild 2020-12-17 kl  18 44 12

HIPPIEKICK:
Building on to what Karin wrote - what needs to be fixed is:
- Your app should show a count of either all todos, or all uncomplete todos (or both)
- Preventing the user to submit empty todos
- Styling issue, see picture in comment

HIPPIEKICK:
Looks like you've done some changes that don't show on your deployed site. Do you have continuous deployment active on Netlify? :)

stjernberg:
I actually can't update to Netlify although I have the continuous deployment active. I can't deploy it manually either.
I attach part of the deploy log. Do you have any idea what could be the problem?

________________________________

stjernberg:
Now the site is updated on Netlify as well!

________________________________

code/src/components/AddTask.js
HIPPIEKICK: Great job with the changes Sara! Just this design bug to be fixed ☝️

PR #136, opened 2020-11-30
To Do List ✅

destinyleia:
Link to my deployed project:

Deploying my site = failed 😞 

I still need to practice how it all works together, but it’s just practice :)

How it all works together!


HIPPIEKICK (changes requested):
Prevent the user from submitting empty todos!

destinyleia:
How do I do that? :)

HIPPIEKICK:
> How do I do that? :)

You can use the `required` attribute or validate in some other way so that the value is not empty before submitting :)

destinyleia:
> > How do I do that? :)
> 
> You can use the `required` attribute or validate in some other way so that the value is not empty before submitting :)

thank you! Think I solved it :)

HIPPIEKICK (approved):

PR #140, opened 2020-12-02
Week 15: todos

monkeybusiness101:
Link to my deployed project:

https://frosty-montalcini-b50721.netlify.app


code/src/components/AddTodo.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react'
+import { useDispatch } from 'react-redux'
+import styled from 'styled-components'
+import Button from '@material-ui/core/Button'
```
HenrikeW, line 4: Cool that you're using material-ui. Looks great!

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+`
+const StyledTextField = styled(TextField)`
+  width: 100%;
+`
+
+export const AddTodo = () => {
+  const dispatch = useDispatch()
+  const [ text, setAddItem ] = useState('')
+  const [ id, setaddId ] = useState(0)
```
HenrikeW: I think you should stick to the conventions when it comes to naming state variables. In this case it would be `[text, setText] `and` [id, setId]`. Otherwise it becomes hard to see that e.g. `text `and `setAddItem `belong together.

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const handleSubmit = (event) => {
+    event.preventDefault()
+    dispatch(
+      tasks.actions.addItem({
+        id, 
+        text,
+      })
+    )
+    setaddId(id + 1)
```
HenrikeW: This logic for the id works for now but it doesn't seem very sustainable because it depends on the component not being rerendered. As soon as you wanted to e.g. store your todos for a longer time, it wouldn't work anymore because it would always add the id 1 as soon as the page is reloaded. 
An alternative could be something like this, either here (then you need to use useSelector to get the information from the store) or inside your reducer (then you would need to adjust its functionality a little): 
`id: store.tasks.length === 0 ? 0 : Math.max(...store.tasks.map(item => item.id)) + 1,`
It looks for the largest id in your item-array and makes the id for the new item 1 larger than it. If the array is empty, the new id will be 0.

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,52 @@
...
+    setAddItem('')
+  }
+
+  return (
+    <Form onSubmit={handleSubmit}>
+      <StyledTextField type="text" value={text} placeholder="Insert task here.." onChange={(event) => setAddItem(event.target.value)} variant="outlined" />
```
HenrikeW: Another indentation with every attribute in a new line would make reading easier.

code/src/components/Item.js
```diff
@@ -0,0 +1,43 @@
...
+  align-items: center;
+  justify-content: space-between;
+  margin: 5px;
+`
+
+export const Item = ({ item }) => {
+  const dispatch = useDispatch()
+  const [check, setCheck] = useState(false)
+  
+
```
HenrikeW: Very much space here :)

code/src/components/Item.js
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <StyledCard>
+      {item.text}
+      <div>
+        <Checkbox type="checkbox" id="button" name="button" value="button" checked={check} onChange={() =>setCheck(prev => !prev)}/>
```
HenrikeW: Same here with the indentation

code/src/components/TotalTodos.js
```diff
@@ -0,0 +1,35 @@
...
+  background-color: tomato;
+  color: white;
+  font-size: 24px;
+`
+const Title = styled.div`
+
+`
+const Total = styled.p`
+
+`
```
HenrikeW: I don't think you need these. If you don't style them, you could just have a div and a p-tag in your return statement

code/src/components/TotalTodos.js
```diff
@@ -0,0 +1,35 @@
...
+
+export const TotalTodos = () => {
+  const totalTodos = useSelector((store) => store.tasks.items.length)
+
+return (
+  <Container>
+    <Image src="https://www.svgrepo.com/show/263286/frog.svg" alt="image toad"/>
```
HenrikeW, line 24: I love the toad :)

code/src/components/Item.js
```diff
@@ -0,0 +1,43 @@
...
+  height: 42px;
+  width: 42px;
+`
+
+const StyledCard = styled(Card)`
+  display: flex;
+  width: 100%;
+  align-items: center;
+  justify-content: space-between;
+  margin: 5px;
```
HenrikeW, line 22: The text inside the card sticks to the left edge of the card. Could need some padding.

code/src/components/TotalTodos.js
```diff
@@ -0,0 +1,35 @@
...
+    <Title>toado list</Title>
+    <Total>{totalTodos} left</Total>
```
HenrikeW: This might be a bit picky. Talking from a user perspective: When I read "3 left" I think "3 left to do" and I would expect the number to change as soon as I mark a todo as completed. In fact, the number only changes if I delete a todo. So something like "Currently 3 todos" or just "3 todos" would be more accurate from my point of view.

HenrikeW (approved):
Great job, Max! Your code is really clean and (nearly everywhere) easy to read. You keep it simple and reached the blue requirements, I believe. I made some comments about naming and indentation, but that's it. And by the way, I love the motto of your page 😄

code/src/components/AddTodo.js
monkeybusiness101: Done

code/src/components/AddTodo.js
monkeybusiness101: Thank you for the feedback. I added a math floor with math random * 1024

code/src/components/AddTodo.js
monkeybusiness101: I will leave it for now, I have a harder time reading it that way. Thank you for pointing it out!

code/src/components/Item.js
monkeybusiness101: I cant see what is wrong with the indentation? :(

monkeybusiness101:
@HenrikeW Thank you for putting the time and providing feedback!!

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,18 @@
...
+    addItem: (state, action) => {
+      const addItems = [...state.items,action.payload]
+      state.items = addItems
+    },
+    removeItem: (state, action) => {
+      const filterItems = state.items.filter((item) => item.id !== action.payload.id)
+      state.items = filterItems
+    }
+  }
+})
```
HIPPIEKICK: Nice job Max! One action (and property of the items) missing though - toggling whether it's a checked item or not. If you check the first todo, create a new one and then remove the checked one - then the second todo gets checked. Also - don't forget to prevent the user from submitting empty todos :)

monkeybusiness101:
@HIPPIEKICK Hi! 

I have added required to the input field
I created a new reducer that changes the required boolean on a specific checkbox. I am experiencing the same issue where the state is correct but the frontend is not. 

Can you please provide me some guidance on how I can solve this issue?

/Max

monkeybusiness101:
@HIPPIEKICK I have made a change so that the correct checkbox is checked or not. Can you please have a look?

HIPPIEKICK:
Looks good Max, approving! 👍  Isn't the local state redundant now though?

monkeybusiness101:
Hi! I also changed now so that I take the proper value from usestate with the use of useeffect instead of passing !check in the dispath

but why is local state redundant? How should I keep track if the checkmark is checked or not?

code/src/components/Item.js
```diff
@@ -0,0 +1,53 @@
...
+        checked: check
+      })
+    )
+}, [check])
+
+  return (
+    <StyledCard>
+     <TextContainer>{item.text}</TextContainer> 
+      <div>
+        <Checkbox type="checkbox" checked={item.checked} onChange={() => { setCheck(prev => !prev) }}/>
```
HIPPIEKICK, line 46: To answer your question, I would use the global state all the way:
```
const handleOnChange = () => {
  dispatch(tasks.actions.changeChecked({ id: item.id, checked: check }))
}

<Checkbox
  checked={item.checked}
  type="checkbox" 
  onChange={() => { handleOnChange}/>
```

PR #150, opened 2021-04-23
Rasmus Persson

skossan:
Link to my deployed project: https://todo-rrpersson.netlify.app/

My reflections on how this week's project turned out: My design skills aren't top notch but I enjoyed building this project from a technical perspective.

### DIFF
--- a/code/src/components/Header.js
+++ b/code/src/components/Header.js
@@ -1,17 +1,26 @@
 import React from "react";
-import { useSelector } from "react-redux";
+import { useSelector, useDispatch } from "react-redux";
 
+import todos from "../reducers/todos";
 import "./HeaderStyle.css";
 
 const Header = () => {
   const items = useSelector((store) => store.todos.items.length);
+  const dispatch = useDispatch();
 
   return (
     <div className="header-container">
       <div className="header-section">
         <div className="header-section-content">
           <p>{items} Todo</p>
-          <button className="header-button">Clear all</button>
+          <button
+            className="header-button"
+            onClick={() => {
+              dispatch(todos.actions.removeAllItems());
+            }}
+          >
+            Clear all
+          </button>
         </div>
       </div>
     </div>
--- a/code/src/components/Header.js
+++ b/code/src/components/Header.js
@@ -1,10 +1,14 @@
 import React from "react";
 import { useSelector } from "react-redux";
 
+import "./HeaderStyle.css";
+
 const Header = () => {
   const items = useSelector((store) => store.todos.items.length);
 
-  return <div>You currently have {items} tasks</div>;
+  return (
+    <div className="header-container">You currently have {items} tasks</div>
+  );
 };
 
 export default Header;
--- a/code/src/components/Header.js
+++ b/code/src/components/Header.js
@@ -9,7 +9,7 @@ const Header = () => {
   return (
     <div className="header-container">
       <div className="header-section">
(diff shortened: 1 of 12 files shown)


code/src/components/Header.js
```diff
@@ -0,0 +1,30 @@
...
+import "./HeaderStyle.css";
+
+const Header = () => {
+  const items = useSelector((store) => store.todos.items.length);
+  const dispatch = useDispatch();
+
+  return (
+    <div className="header-container">
+      <div className="header-section">
+        <div className="header-section-content">
```
Ann-Sofi, line 14: Good naming of classes 👍

code/src/components/TodoList.js
```diff
@@ -0,0 +1,50 @@
...
+                onChange={() => {
+                  dispatch(todos.actions.toggleComplete(todo.id));
+                }}
+              />
+
+              <p
+                className="todo-list-item-text"
+                onClick={() => {
+                  dispatch(todos.actions.toggleComplete(todo.id));
+                }}
```
Ann-Sofi: The toggle works like a charm, nice work! 👍 One thing to think about if you want to make it a bit more user-friendly. If I write a short text, like "öl"  for example, it gets a bit tricky to toggle beween complete and not complete because the toggle is on the word and the word is very short and small.  To make it a bit easier for the user you could do so the entire section is "clickable",

code/src/components/TodoList.js
```diff
@@ -0,0 +1,50 @@
...
+
+import todos from "../reducers/todos";
+import "./TodoListStyle.css";
+
+const TodoList = () => {
+  const dispatch = useDispatch();
+  const items = useSelector((store) => store.todos.items);
+
+  return (
+    <div>
```
Ann-Sofi, line 13: I realised I have just the same issue as you. When I write something long, it does not fit inside the task-section. It goes way beyond the intended width and messes up the layout. What we want is for the text to line-break.. Let me know if you have a good solution for that!!

code/src/components/NewTask.js
```diff
@@ -0,0 +1,52 @@
...
+  const handleChange = (event) => {
+    setInputValue(event.target.value);
+  };
+
+  let newTodo;
+
+  const handleOnClick = (event) => {
+    event.preventDefault();
+    if (inputValue === "") {
+      alert("You can't add an empty todo.");
```
Ann-Sofi: Great that you made so the user can't input an empty task. When the alert has pop-ed up, the page is blank.. so a suggestion could be to take the user back to the app to try again instead of the blank page. :-)

code/src/components/NewTask.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from "react";
+import { useDispatch } from "react-redux";
+import uniqid from "uniqid";
+
```
Ann-Sofi, line 4: Nice that you follow the guidelines and put a empty line beween different imports :-)

Ann-Sofi (approved):
Great job! ✨ 
Nice structure and the code is clean and easy to read. I tried it on both mobile and in browser. The app works like it should and it's easy to know how to use it. If you want the published site to be more personal, don't forget to change title in index.html 😃

KarolCislowski (changes requested):
please fixe all red flags risen by github-linter

KarolCislowski (approved):
Great APP

PR #161, opened 2021-04-24
Project TODOs Estefanía

esteficodes:
Link to my deployed project:

https://todos-app-estefania.netlify.app/

I think this week's project went overall very good despite the lack of time I had. This is why I'm handing in a version that I plan to improve in the upcoming days, as it still needs a lot of work. I'm happy about having learned how Redux works a little bit more in detail and also about my first time using both styled-components and regular css. 

I would like to get more instruction on how is the most convenient way of dividing the project into components, as I'm a little bit lost in this area.


code/src/components/Header.js
```diff
@@ -0,0 +1,25 @@
...
+`
+
+//This component shows the app header and the tasks to be completed
+//Accesses the global state and gets the complete current to do list. Then, filters out the
+//tasks in that list with property complete: false and returns the length of that array
+ const Header = () => {
+  return (
+    <div className="header">
+      <h1 className="app-title">MY TODOs APP</h1>
+      <CurrentDate>{moment(Date.now()).format('MMMM Do YYYY, h:mm:ss a')}</CurrentDate>
```
honkoan, line 18: I like that you have the date and time! 🙌🏻  I'm not sure if I remember correctly, but I think moment is not the best alternative to use because they are not updating it anymore.

code/src/components/TodoList.js
```diff
@@ -0,0 +1,41 @@
...
+              {items.map((todo, index) => (
+                  <div key={todo.id} className="todo-item">
+                     <p>{todo.description}</p>
+                     <input
+                     type="checkbox"
+                     checked={todo.isComplete}
+                     onChange={() => dispatch(todos.actions.toggleComplete(todo.id))}
+                   />
+                <span className={todo.isComplete === true ? 'todo-description-done' : 'todo-description'} key={todo.id}>{todo.description}</span>
+                <Timestamp>added {moment(todo.createdAt).startOf('hour').fromNow()}</Timestamp>
```
honkoan, line 30: I love that you did this timestamp! 🙌🏻 I will add it to my own project if I have time later (which I never have 🤣). 
Now it shows the same timestamp for all todos, so something needs to be fixed. But it is awesome that you tried to do all these things even though you had the lack of time! 😍 I did just the basic things.

code/src/index.css
```diff
@@ -1,13 +1,77 @@
...
+  font-weight:100;
+  margin-top: 8px;
+  cursor: pointer;
+  transition: all 150ms;
+  margin-top: 26px;
+}
+
+.button:hover {
+    background-color: #808080; 
+}
```
honkoan, line 77: Such a beautiful page with just this little amount of CSS! 🌟 Well done, Estefania! 

Your css file is also easy to understand becouse of good class naming.

code/src/components/TodoList.js
```diff
@@ -0,0 +1,41 @@
...
+        <div className="list-container">
+              {items.map((todo, index) => (
+                  <div key={todo.id} className="todo-item">
+                     <p>{todo.description}</p>
+                     <input
+                     type="checkbox"
+                     checked={todo.isComplete}
+                     onChange={() => dispatch(todos.actions.toggleComplete(todo.id))}
+                   />
+                <span className={todo.isComplete === true ? 'todo-description-done' : 'todo-description'} key={todo.id}>{todo.description}</span>
```
honkoan, line 29: Oh this is cool! It changes the class name depending on if the todo is done or not. I also want to add this feature to my project.🙂 You are probably planning to style it somehow later? I read from your pull request that you are going to continue working on this project. Looking forward to seeing what your app will look like when you're done because I like it already. 💐

code/src/components/TodoList.js
```diff
@@ -0,0 +1,41 @@
...
+const TodoList = () => {
+    const items = useSelector((store) => store.todos.items);
+   
+    const dispatch = useDispatch();
+
+    return(
+        <div className="list-container">
+              {items.map((todo, index) => (
+                  <div key={todo.id} className="todo-item">
+                     <p>{todo.description}</p>
```
honkoan, line 23: The todos are shown twice, was it your purpose? If it wasn't, you propably don't need this p-tag with {todo.description}, because the span-tag is also showing todo.description.

code/src/components/Header.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import styled from 'styled-components';
+import moment from 'moment';
+
+
+const CurrentDate = styled.p`
+  margin: 0 0 15px 0;
+  color: white;
+`
+
```
honkoan, line 10: You also used the styled components, nice! 💅🏻

honkoan (changes requested):
Your site is looking awesome and it made me smile when I saw it! 😊 I love those colors and that yellow button! 🌈 There were lot's of things that inspired me how I could improve my own todo app! Really nice work, Estefania. 🥰 

It was really nice to read your code which is clear and easy to understand. I noticed that you are planning to continue working on this project so I didn't pay attention to for example console.logs. 

In my opinion you divided your project into components in a clear way. I did also a component for "todo item", but I think it is not necessary because there isn't so much code in TodoList -component.  

The delete button is not working properly. It is not removing the correct todo. That is why I am requesting changes. The app should also show a count of todos. 

I tried to figure out what is the problem with delete button, but I couldn't. I'm really interested in what is the solution for that, so if you can't figure it out maybe we could try to solve it with the team. 🦊 

Your app will be really good after you have done changes because it is so nice already. 🙌🏻 😃 Please ask if there is something strange in my comments or something you don’t understand.

code/src/reducers/todos.js
```diff
@@ -0,0 +1,62 @@
...
+            // v1 - Immutability
+            // const decreasedItems = store.items.filter(todo => todo.id !== action.payload);
+
+            // store.items = decreasedItems;
+
+            //v2 - Mutability
+            store.items.splice(action.payload, 1);
+        },
+        addTodo: (store, action) => {
+            console.log(action);
```
KarolCislowski: console.log 😭

code/src/reducers/todos.js
```diff
@@ -0,0 +1,62 @@
...
+           // })
+        },
+        removeTodo: (store, action) => {
+            // v1 - Immutability
+            // const decreasedItems = store.items.filter(todo => todo.id !== action.payload);
+
+            // store.items = decreasedItems;
+
+            //v2 - Mutability
+            store.items.splice(action.payload, 1);
```
KarolCislowski: you are passing the id as the payload, not the index, that is why your reducer removes the wrong item from the array.

You have 3 options here(from best to worst):
1. use the Immutability version because it looks good for me,
2. find index of task with id from the payload, something like this:
```jsx
const index = store.map(e => e.id).indexOf(action.payload);
store.items.splice(index, 1);
```
3: pass index as payload.

code/src/components/TodoForm.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    return(
+        <>
+        <form className="form" onSubmit={onFormSubmit}>
+            <input className="input"
+               placeholder="Write your todo here..."
+               type="text"
+               value={value}
+               onChange={e => setValue(e.target.value)}
```
KarolCislowski, line 31: FIGHT WITH LAZYNESS!! 🔫 add REQUIRED to make your to-do free from empty tasks!! 🎖️ 😆

code/src/components/TodoForm.js
```diff
@@ -0,0 +1,40 @@
...
+            description: value,
+            isComplete: false
+        };
+
+        dispatch(todos.actions.addTodo(newTodo));
+        setValue('');
+    }
+
+    return(
+        <>
```
KarolCislowski: You have only one element to return (`<form>`) you shouldn't use <></> here 🤓

KarolCislowski (changes requested):
I left tip on how to fix your reducer so do it asap 🏁 

Dividing your app into components is not an easy thing to present as a short text in code review. Please try to book a 1:1 session. 🤓

KarolCislowski (approved):
Great work 🥳

PR #170, opened 2021-04-25
Ylvas Todo app

sssylvasss:
Link to my deployed project:
https://stuff-todo.netlify.app/


code/src/components/TodoItem.js
```diff
@@ -0,0 +1,79 @@
...
+import { todos } from "../reducers/todos";
+
+
+export const TodoItem =({ todo }) => {
+    const dispatch = useDispatch()
+
+    return (
+        <Main>
+            <TodoContainer>
+                <div
```
jesinkey, line 17: Nice solution with the div as the "styled button". never would have thought of that.

code/src/components/NoTodo.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
```
jesinkey: Nice job on having a component that mounts even when todos are empty.

code/src/components/TodoItem.js
```diff
@@ -0,0 +1,79 @@
...
+    color: #BDBDBD;
+    padding-top: 7px;
+    margin-right: 10px;
+        @media (min-width: 768px) {
+        font-size: 12px;
+        padding-top: 5px;
+    }
+`;
+
+const Span = styled.p`
```
jesinkey: Maybe a different name than span? I thought it would be styled differently? But thats just semantics and maybe i'm wrong?

jesinkey (approved):
Hej Ylva =)
I like that you used style components! It was something I tried but still haven't wrapped my head around. 
I saw something with your cute hamsterwheel image, when you hover over it a white square appears in the top right corner. I couldn't see what it was due to and couldn't see it in the elements page in the inspector?
Otherwise great project with a great design!

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,105 @@
+import React, { useState } from 'react'
+import { useDispatch } from 'react-redux'
```
KarolCislowski, line 2: I think you need to configure your Prettier to add semicolons for you because you are not consistent about it and it is should be automated 😸

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,105 @@
...
+
+export const AddTodo = () => {
+    const [inputValue, setInputValue] = useState(''); 
+    const dispatch = useDispatch();
+ 
+    const onItemAdd = (event) => {
+        event.preventDefault();
+        dispatch(
+            todos.actions.addTodoItem({
+                newTodo: {
```
KarolCislowski, line 18: Another thing you should check in your Prettier's settings is to set indentation for 2 spaces by default.

KarolCislowski (approved):
You built a great app and the hamster animation is so sweet I gonna die! 😻

PR #176, opened 2021-04-25
Todo List Ulysses Veghelyi 

ghost:
Link to my deployed project:
https://dazzling-swirles-b0f09e.netlify.app/ 

This week project turned out good, always better than expected compered to what you are feeling on mondays. 

### DIFF
--- a/code/src/components/ClearAllButton.js
+++ b/code/src/components/ClearAllButton.js
@@ -1,0 +1,44 @@
+import React from "react";
+import { useDispatch } from "react-redux";
+import styled from "styled-components";
+
+import  { todos }  from "../reducers/todos";
+
+const Buttons = styled.div `
+    display: flex;
+    justify-content: center;
+    margin: 10px;
+`;
+
+const DeleteAllTodo = styled.button `
+    display: flex;
+    justify-content: center;
+    background:linear-gradient(25deg, #2BC0E4 0%, #EAECC6 100%);
+    display:inline-block;
+    padding:0.35em 1.2em;
+    border:1px solid #FFFFFF;
+    margin-top 20px;
+    border-radius:5px;
+    box-sizing: border-box;
+    text-decoration:none;
+    font-family:'Roboto',sans-serif;
+    font-weight:500;
+    color:#fff;
+    text-align:center;
+`;
+
+export const ClearAllButton = () => {
+  const dispatch = useDispatch();
+  
+  const clearAllTodos = (event) => {
+    dispatch(todos.actions.clearAll());
+  };
+
+  return (
+    <Buttons>
+        <DeleteAllTodo onClick={clearAllTodos}>
+            Delete All
+        </DeleteAllTodo>
+    </Buttons>
+  )
+}
--- a/code/src/components/TodoInput.js
+++ b/code/src/components/TodoInput.js
@@ -1,0 +1,65 @@
+import React, { useState } from "react";
+import { useDispatch } from "react-redux";
+import styled from "styled-components";
+
+import { todos }   from "../reducers/todos.js";
+
+const FormContainer = styled.form `
+    padding: 20px;
+    text-align: center;
+`;
+
+const FormInput = styled.input `
+    padding: 10px 0px 5px 3px;
+    font-size: 16px;
+    border-radius: 4px;
+    position: relative;
+    border: 1px solid #fff;
+`;
+
+const AddButton = styled.input `
+    background:transparent;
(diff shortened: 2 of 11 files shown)


code/src/components/TodoList.js
```diff
@@ -0,0 +1,41 @@
...
+  width: 95%;
+  padding: 5px;
+  opacity: 0.8;
+  @media (min-width: 667px) {
+    width: 50%;
+  }
+  @media (min-width: 1024px) {
+    width: 350px;
+  }
+`;
```
sarianne-peltola, line 24: When I add some short named to-do like "hi" on the page, it will be added inline the previous item (if there is space). So the tasks will not be as a list, but rather on one row. For that I was thinking maybe you could use flexbox here. Something like 
```javascript
display: flex;
flex-direction: column;
align-items: flex-start;
```

Then you might want to do a 'aling-self: flex-end' to the ToDoCompleted, so the counter is on the right side. If it works that way in the styled components :)

code/src/reducers/todos.js
```diff
@@ -0,0 +1,38 @@
...
+
+const initialState = {
+  list: {
+    name: "Todo List",
+    items: [
+      { id: 1, text: "Hi and Welcome to this Todo list!😄", isComplete: false },
+      { id: 2, text: "You can add your own todo⬆️", isComplete: false },
+      { id: 3, text: "Check them if you are done✔️", isComplete: false },
+      { id: 4, text: "See how many tasks are done🔢", isComplete: false },
+      { id: 5, text: "Please delete these now!🤓", isComplete: false },
```
sarianne-peltola, line 11: Quite nice actually so you notice all the implemented features :)

code/src/components/TodoInput.js
```diff
@@ -0,0 +1,65 @@
...
+    setInputText("");
+  }
+
+  return (
+    <FormContainer onSubmit={handleSubmit}>
+      <FormInput
+        type="text"
+        onChange={(event) => setInputText(event.target.value)}
+        value={inputText}
+        required
```
sarianne-peltola, line 60: Good add the "required", forgot to add it myself!

code/src/components/TodoInput.js
```diff
@@ -0,0 +1,65 @@
...
+    border-radius:4px;
+    margin:10px;
+    box-sizing: border-box;
+    text-decoration:none;
+    text-transform:uppercase;
+    font-family:'Roboto',sans-serif;
+    font-weight:400;
+    color:#fff;
+    text-align:center;
+
```
sarianne-peltola, line 33: Just some extra line here, but who cares right :)

sarianne-peltola (approved):
Nice work especially with the styled components! Code is super clean, there is nothing really to make marks on and definitely satisfies the requirements for the week. One thing I just mentioned about the TodoList that you might want to check if you see the same issue with the list of tasks. Styling of the app is nice with some cool emojis. I liked that you had left some ready-made "todos" to show what features you have implemented in the app. Really good job! It was a pleasure to read your code and I got even some tips on my code too :)

code/src/components/TodoInput.js
KarolCislowski: I care 😆

code/src/components/ClearAllButton.js
```diff
@@ -0,0 +1,44 @@
...
+    margin: 10px;
+`;
+
+const DeleteAllTodo = styled.button `
+    display: flex;
+    justify-content: center;
+    background:linear-gradient(25deg, #2BC0E4 0%, #EAECC6 100%);
+    display:inline-block;
+    padding:0.35em 1.2em;
+    border:1px solid #FFFFFF;
```
KarolCislowski, line 19: in line 26 you used short notation, why not here??

code/src/components/ClearAllButton.js
```diff
@@ -0,0 +1,44 @@
...
+  
+  const clearAllTodos = (event) => {
+    dispatch(todos.actions.clearAll());
+  };
+
+  return (
+    <Buttons>
+        <DeleteAllTodo onClick={clearAllTodos}>
+            Delete All
+        </DeleteAllTodo>
```
KarolCislowski, line 41: what happened with indentation here? you need to check the Prettier config to be sure you have tab depth set to 2 spaces.

KarolCislowski (approved):
Nice app 😸 
Please check indentation configuration in your Prettier's settings 🤓

PR #182, opened 2021-04-25
Project 15 - Todo List

AsaHildebrand:
Link to my deployed project:

My reflections on how this week's project turned out: I think it turned out well except for a lot of work with the deploying to Netlify

The items array is commented out because I didn't want it to show on the live page but it is useful to show it when testing the code so that is why I kept the items in the array.


code/src/components/AddTask.js
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from 'react'
+import { useDispatch } from 'react-redux'
+import uniqid from 'uniqid'
+import styled from 'styled-components'
+
+import todos from '../reducers/todos'
+
+const FormContainer = styled.form`
```
Polkapetunian, line 8: Great that you have used styled components! =)

code/src/components/AddTask.js
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from 'react'
+import { useDispatch } from 'react-redux'
+import uniqid from 'uniqid'
+import styled from 'styled-components'
+
```
Polkapetunian, line 5: Good that you remembered a gap between external dependencies and reducers.

code/src/components/AddTask.js
```diff
@@ -0,0 +1,61 @@
...
+import todos from '../reducers/todos'
+
+const FormContainer = styled.form`
+  display: flex;
+  justify-content: center;
+  border-bottom: 1px solid grey;
+  background-color: #f0f0fc;
```
Polkapetunian: Colors are written in two different ways

code/src/components/AddTask.js
```diff
@@ -0,0 +1,61 @@
...
+  const dispatch = useDispatch()
+
+  const onFormSubmit = (e) => {
+    e.preventDefault()
+
+    const newTodo = {
+      id: uniqid(),
+      description: value,
+      isComplete: false,
+      dueDate: new Date().toJSON()
```
Polkapetunian: I think dueDate is not a very describing name, since it sets the date when the task is created, not when it is due...(supposed to be completed).

code/src/components/Header.js
```diff
@@ -0,0 +1,54 @@
...
+import styled from 'styled-components'
+
+import todos from '../reducers/todos'
+
+const HeaderContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  background-color: #c7ebde;
+  padding: 15px 0 15px 0;
```
Polkapetunian: This could be written just 15px 0;

code/src/components/Header.js
```diff
@@ -0,0 +1,54 @@
...
+const Header = () => {
+  const items = useSelector((store) => store.todos.items)
+  const completedItems = items.filter(singleTodo => singleTodo.isComplete)
+  
+  const dispatch = useDispatch()
+  
+   return (
+    <HeaderContainer>
+      <Title tabIndex="0">Your Todo List</Title>
+      <Todo tabIndex="0">Completed tasks: {completedItems.length}/{items.length}</Todo>
```
Polkapetunian, line 45: Not sure these should be focusable... MDN says"If an element can be focused using the keyboard, then it should be interactive; that is, the user should be able to do something to it and produce a change of some kind (for example, activating a link or changing an option)."https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard

code/src/components/TaskList.js
```diff
@@ -0,0 +1,86 @@
...
+const TodoDiv = styled.div`
+  display: flex;
+  justify-content: center;
+  width: 100%;
+  border-bottom: 1px solid grey;
+`
+
+const Todo = styled.p`
+  font-size: 18px;
+  padding: 0 5px 0 5px;
```
Polkapetunian: Could be simplified to 0 5px;

code/src/components/TaskList.js
```diff
@@ -0,0 +1,86 @@
...
+    <TodoContainer>
+      {items.map(todo => (
+        <TodoDiv key={todo.id}>
+          <CheckButton
+            type="checkbox"
+            checked={todo.isComplete}
+            onChange={() => dispatch(todos.actions.toggleComplete(todo.id))}
+          />
+          <Todo tabIndex="0">{todo.description}</Todo>
+          <Text tabIndex="0">{new Date(todo.dueDate).toLocaleDateString()}</Text>
```
Polkapetunian: Same here: It probably shouldn't be focusable

Polkapetunian (approved):
A very nice todo app! The design as well as the code is clean. It works as it should and red goals are reached. 🏅 ✔

code/src/components/NoData.js
```diff
@@ -0,0 +1,30 @@
...
+  margin: 60px 40px 0px 40px;
+  color: #032640;
+`
+
+const NoData = () => {
+  const items = useSelector((state) => state.todos.items)
+
+  return (
+    <>
+    {!items.length && (
```
KarolCislowski, line 22: it should be checked in the parent component if it is needed because it would save resources used to mount this component😭

code/src/reducers/todos.js
```diff
@@ -0,0 +1,51 @@
...
+  
+        if (todo.id === action.payload) {
+          
+          return {
+            ...todo,
+            isComplete: !todo.isComplete
+          }
+        
+        } else {
+
```
KarolCislowski, line 25: why so many holes?? 🎯  🔫

KarolCislowski (approved):
Nice app 🦄 Remember to avoid loading components if they will show nothing 😉

PR #202, opened 2021-04-28
project-15-shaun-newmark

aicb1:


johlofg (approved):
Nice project! 
I don´t have that much to add. 
In css you mix some hex, rgb and name for colors otherwise good job. 
Tested the app and it works as it should and at least all blue goals are reached. 
Your code is clear and easy to read.  Good job!

code/src/index.css
```diff
@@ -11,3 +18,40 @@ code {
...
+  display: flex;
+  flex-direction: column;
+  margin: 10px;
+  background: linear-gradient(
+    189deg,
+    rgba(166, 228, 255, 1) 0%,
+    rgba(125, 113, 195, 1) 72%
+  );
+  padding: 30px;
+  width: auto;
```
KarolCislowski, line 32: It is not a good practice to do that. You're completely losing control over your container.
If you still want to use it as that you need to think about making your inputs stretch ass well, or move your content to the center.

code/src/index.css
```diff
@@ -1,3 +1,10 @@
+html {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-color: black;
```
KarolCislowski, line 5: why you used `html` instead of `body` to set these styles?

code/src/components/TodoList.js
```diff
@@ -0,0 +1,28 @@
...
+      {items.map((todo) => (
+        <div key={todo.id} className="todo-item">
+          <button type="button" className="delete-btn" onClick={() => dispatch(todos.actions.removeTodo(todo.id))}>
+            X
+          </button>
+          <p>{todo.description}</p>
+          <input
+            type="checkbox"
+            checked={todo.isComplete}
+            onChange={() => dispatch(todos.actions.toggleComplete(todo.id))} />
```
KarolCislowski, line 21: IMHO, Layout: checkbox, text, delete btn, wóuld be more intuitive. 
To make it even cleaner you can wrap checkbox and text with a container and kick your delete button by using `space-between`

KarolCislowski (approved):
Nice and clean logic. 
Remember that your projects are your ticket to a new career so make them look as nice as possible.

PR #220, opened 2021-11-19
Todo - Madeleine Svensson

madeleinesvensson:
Link to my deployed project:
https://autumntodo.netlify.app

It went good I would say. My imposter syndrome went away and I'm getting reducers now more than last week. I'm starting to like styled components but not really liking them because i'm not really seeing the structure of what's what 

-


code/src/App.js
```diff
@@ -1,9 +1,41 @@
-import React from 'react'
+import React from "react";
+import { Provider } from "react-redux";
+import { combineReducers, createStore } from "@reduxjs/toolkit";
+
+import { todos } from "./reducers/todos";
```
jakobxlindstrom, line 5: Is this one needed in the App, It looks to be not mounted in App?

code/src/components/Buttons.js
```diff
@@ -0,0 +1,45 @@
+import styled from "styled-components/macro";
```
jakobxlindstrom, line 1: This /macro what does it do? 

Otherwise nicely done with styled components!

code/src/components/Footer.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import styled from "styled-components";
+
+const FooterWrapper = styled.div`
```
jakobxlindstrom, line 4: What a great way to make a reusable component. Put this somewhere accessible so that you can reach it for future projects fast and easy. I will take this with me to my future projects.

jakobxlindstrom (commented):
Great looking app Madde. I really like the extra time you put into moving a finished task into a new section in your container. It looks really pro and I think you have shown great skills here! 
Your site looks great on mobile as well as desktop which is important because I do think a todo app will most likely be used on the phone before the computer. 
Keep up the good work!

PR #233, opened 2021-11-20
ToDo-List Birgit

nehrwein:
Link to my deployed project: https://nehrwein-todo-list.netlify.app/

My reflections on how this week's project turned out: I am happy with the result. If I would have more time, I'd like to make the entries editable and add some more information like a due date to them.


code/src/components/AddTask.js
```diff
@@ -0,0 +1,96 @@
...
+  const [category, setCategory] = useState('')
+  const dispatch = useDispatch()
+
+  const onAddToDo = (input, category) => {
+    dispatch(todos.actions.addToDo({ input, category }))
+    setInput('')
+    setCategory('')
+  }
+
+  const options = ['Work', 'Study', 'Shopping', 'Other']
```
DALA746, line 65: Interesting approach to put all the options into an array and then map through that array to show each option. I will definitely take that approach with me for future projects.

code/src/components/AddTask.js
```diff
@@ -0,0 +1,96 @@
...
+        <option disabled value="">Category</option>
+        {options.map((opt)=> (
+          <option 
+          key={opt} 
+          value={opt}>{opt}
+        </option>))}
+      </select>
+      <AddBtn 
+        type='submit'
+        disabled = {!input}
```
DALA746, line 88: I wrote `disabled={input === ""}`, so if the input is empty the button should be disabled. But here you write `disabled = {!input}`, I guess it says if it is not input then do disabled button or if input false disabled button?

code/src/components/CompletedToDos.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import { useNavigate } from 'react-router'
```
DALA746, line 3: Did useNavigate solve your problem with the pages that you had? What does it do? Navigate to the page that you want to go to?

code/src/components/ToDos.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react'
+import { useSelector } from 'react-redux'
+import { useDispatch } from 'react-redux'
+import styled from 'styled-components/macro'
+import moment from 'moment'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faTrashAlt } from '@fortawesome/free-solid-svg-icons'
```
DALA746, line 6: I did not think that there is a package on npm for font awesome. I put a link in public index.html and then used `<i> `tags for icons. But I think this approach is more relevant for react, because react has a lot of packages and it is better to be consistent and use the same for everything, like in this case package for font awesome.

code/src/components/ToDos.js
```diff
@@ -0,0 +1,109 @@
...
+  margin-bottom: 10px;
+  padding: 5px;
+  font-size: 20px;
+  border: 1px solid var(--red);
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  border-radius: 5px;
+  opacity: ${props => props.completed ? 0.3 : 1};
+  background-color: ${props => props.category === 'Study' ? 'var(--lax)' : props.category === 'Work' ? '#6997b0' : props.category === 'Shopping' ? '#9e9dd9' : '#facdc5'};
```
DALA746, line 42: That one was gold! I really love styled-components now and what they can do!

code/src/index.css
```diff
@@ -1,13 +1,23 @@
+@import url('https://fonts.googleapis.com/css2?family=Quicksand:wght@300;400;500;600;700&display=swap');
+
+:root {
+  --lax: #e59089;
+  --white: #fff4f3;
+  --brown: #9b0c33;
+}
```
DALA746, line 7: I have never used that approach, what does it do? Setting root colors to the page?

code/src/reducers/todos.js
```diff
@@ -0,0 +1,35 @@
...
+        category: action.payload.category
+      })
+    },
+    removeToDo: (state, action) => {
+      state.items = state.items.filter((item) => item.id !== action.payload)
+    },
+    toggleIsComplete: (state, action) => {
+      const task = state.items.find((item) => item.id === action.payload)
+      task.isComplete = !task.isComplete
+      state.items = state.items.sort((a,b) => a.isComplete - b.isComplete)
```
DALA746, line 30: Can you maybe leave comments about your functions and what they do? I see that you filtering items array, and then you are setting task from true to false on line 26, is that correct? Then you doing sorting on items array with task that is false, or ?

code/src/reducers/todos.js
```diff
@@ -0,0 +1,35 @@
...
+const initialState = {
+  items: []
+}
+
+const todos = createSlice({
+  name: 'todos',
+  initialState,
+  reducers: {
+    addToDo: (state, action) => {
+      state.items.unshift({
```
DALA746, line 14: This `unshift` method returns an array as reverse?

DALA746 (commented):
Clean and readable code! I really like your approach with options, to put them in an array. Can you maybe leave more comments about your code, especially about functions? In that case, you explain it more in a human language :) Really love your styling, regarding different colors on your todos, I don't really know which color is for work or studies, maybe put a text about the category in the todo?  

I really enjoyed going through your code and I learned a lot of new approaches. Nice job and I am looking forward to see more of your wonderful projects. 🥇

code/src/components/AddTask.js
nehrwein: It means the same as `disabled={input === ""}`. If there is no input, then the button will be disabled

code/src/components/CompletedToDos.js
nehrwein: useNavigate is just another version of useHistory. I guess, navigate is a newer approach than history and to me it also makes more sense to call this method navigate. 
But it has nothign to do with the routes problem I had. This was solved by making an addition to the netlify.tomI

code/src/components/ToDos.js
nehrwein: I found this approach on the [Fontawesome-pages](https://fontawesome.com/v5.15/how-to-use/on-the-web/using-with/react) 🙂

code/src/components/ToDos.js
nehrwein: I am looking forward to see some cool styling ideas in your future projects! 😺

code/src/index.css
nehrwein: I am defining there some colors that I am using more often on my page. I can give for example a button, a div and paragraph the color: var(--lax). When I then later on decide, that I want to change the color to another value, I do this just on time in the :root and all the elements get the new coloring immediately

code/src/reducers/todos.js
nehrwein: Sorry, you are right, I should be better with the commenting!
Line 26 has nothing to do with Line 27, except that they are both happening, when the user clicks the checkbox. 
So line 25/26 just take care of the store getting updated every time, that the user clicks the checkbox of a task.
Line 27 sorts the array of items by the key "isComplete", so that the completed items move to the end

code/src/reducers/todos.js
nehrwein: no, it just adds a new element to the beginning of an array. (push adds new elements to the end)

nehrwein:
Thank you so much for your code review. Great feedback to add a color-scheme to the app and more comments to my code!
🙂

PR #236, opened 2021-11-20
Jessica Nordahls TO-DO app

JessicaNordahl:
Link to my deployed project:
https://to-do-app-jessica-nordahl.netlify.app/

I am very proud of my choise of "clean" design to this weeks project. 
I learned to work with styled components more in depth and that acctually workes very well with the flow of Redux and the usage of React's hooks.

### DIFF
--- a/code/src/components/AddTodo.js
+++ b/code/src/components/AddTodo.js
@@ -2,7 +2,7 @@ import React, { useState } from "react";
 import { useDispatch } from "react-redux";
 import styled from 'styled-components';
 
-import Todos from '../reducers/todos'
+import Todos from '../reducers/Todos'
 
 //styling the component here:
 const Container  = styled.div`
--- a/code/src/components/AddTodo.js
+++ b/code/src/components/AddTodo.js
@@ -4,63 +4,59 @@ import styled from 'styled-components';
 
 import Todos from '../reducers/Todos'
 
-//styling for the component here:
+//styling the component here:
 const Container  = styled.div`
-    display: flex;
-    flex-direction: column; 
-    align-items: center;
-    justify-content: space-evenly;
-    width: 90%;
-    padding: 20px;    
-    margin: 70px 50px;
-    border: 2px solid grey;
-    overflow: scroll;
-    `
-const Label = styled.label`
-    display: flex;
-    border-radius: 3px;
-    padding: 0.5rem 0;
-    margin: 0.5rem 1rem;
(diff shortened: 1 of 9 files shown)


code/src/components/AddTodo.js
```diff
@@ -0,0 +1,76 @@
...
+
+import Todos from '../reducers/Todos'
+
+//styling the component here:
+const Container  = styled.div`
+display: flex;
+flex-direction: block; 
+align-items: start;
+justify-content: flex-start;
+width: 350px;
```
KaraHowes, line 13: It's a personal thing, but I generally set widths as %; maybe with setting a max- and/or a min-width. That way it ensures your to-do  list can be viewed on all devices :-)

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,76 @@
...
+    transition: 0.5s;
+    color: #white;
+}
+`
+
+const AddTodo = () => {
+    const [input, setInput] = useState('');
+    const dispatch = useDispatch();
+    const onAddTodo = () => {
+        dispatch(Todos.actions.addTodo(input)) //by passing the input here, we know what is the current property from the user
```
KaraHowes, line 59: If you wanted to clear the input after you have added the to-do, you could add 'setInput('')' at the end of the dispatch on line 59.......so it would read 'dispatch(Todos.actions.addTodo(input)); setInput('');'

code/src/components/TodoList.js
```diff
@@ -0,0 +1,97 @@
...
+`
+const Display = styled.div`
+display: flex;
+width: 100%
+flex-direction: flex-start;
+align-items: center;
+`
+const List = styled.h3`
+font-family: 'Hind Vadodara', sans-serif;
+`
```
KaraHowes, line 53: I think you use this font through-out the project, in which case you could just declare it globally in the index.css :-)

KaraHowes (commented):
Great Job Jessica! As you said in your comment, your to-do list is very clean and with an elegant aesthetic. You should be proud! Especially since you did all of the styling using styled components. Having said that, I think for things which are constant through-out (such as font-family), you could declare this in the global state. 
Your code is easy to read and well-explained. I have included a few tips that you could implement, such as how to clear the text input (which gives a nicer user-experience and is easy to implement).
Overall I am extremely impressed and you should be happy. Great job!

PR #239, opened 2021-11-21
Zancotti todo app

Zancotti:
Link to my deployed project:
https://zancotti-todoapp.netlify.app/


code/public/index.html
```diff
@@ -1,34 +1,13 @@
 <!DOCTYPE html>
 <html lang="en">
-
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <!--
-      Notice the use of %PUBLIC_URL% in the tags above.
-      It will be replaced with the URL of the `public` folder during the build.
-      Only files inside the `public` folder can be referenced from the HTML.
```
waliem, line 10: A suggestion would be to add a og-tags and links here.

code/src/components/CategoryIcon.js
```diff
@@ -0,0 +1,78 @@
+import React from "react";
+import styled from "styled-components/macro";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import {
+  faWavePulse,
```
waliem, line 5: I really like that you used descriptive icons for your buttons.

code/src/components/Header.js
```diff
@@ -0,0 +1,164 @@
...
+    } else {
+      return Math.round((completedList.length / todoList.length) * 100);
+    }
+  };
+
+  return (
+    <HeaderContainer>
+      <TitleDateContainer>
+        <Title>Your Todo's</Title>
+        <DateContainer>{moment(date).format(`MMM D, YYYY`)}</DateContainer>
```
waliem, line 29: curious, why does month have three M here?

code/src/components/TodoList/AddButton.js
```diff
@@ -0,0 +1,42 @@
...
+  align-items: center;
+  background-color: ${mainColor};
+  border-radius: 50%;
+  border: lightgrey 1px solid;
+  margin: 20px;
+  font-size: 30px;
+  cursor: pointer;
+  &:hover {
+    background-color: rgba(217, 138, 96, 0.8);
+    transition: all 200ms ease-in-out;
```
waliem, line 40: is this working, I tried to se the effects but I might be targeting the wrong div in the dev tools.

code/src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,131 @@
...
+    (todo) => todo.category === "activity" && !todo.completed
+  );
+  const business = todos.filter(
+    (todo) => todo.category === "business" && !todo.completed
+  );
+  const family = todos.filter(
+    (todo) => todo.category === "family" && !todo.completed
+  );
+  const completed = todos.filter((todo) => todo.completed);
+
```
waliem, line 29: This was a nice solution to the rendering of categories and completed and uncompleted 👍

code/src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,131 @@
...
+  // Using a useEffect to recognise changes in the global state and then
+  // either scrollIntoView or not.
+  useEffect(() => {
+    if (scrollIntoView) {
+      ButtonsContainerRef.current.scrollIntoView();
+    }
```
waliem: Curious, what does this scrollintoview actually do? I couldn't locate it when using the app.

code/src/components/TodoList/TodoListItem.js
```diff
@@ -0,0 +1,99 @@
...
+        <ItemLocation>{item.place}</ItemLocation>
+        <ItemDueDate style={{ color: dueDateColor() }}>
+          Due date {moment(new Date(item.dueDate)).format(`MM/DD-YYYY`)}
+        </ItemDueDate>
+        <CreatedAt>
+          Created {moment(new Date(item.createdAt)).fromNow()}
+        </CreatedAt>
+      </div>
+      <TrashIcon>
+        {/* When pressing the trashcan button the dispatch call for the removeTodo function in the global state. The payload is the id of the todo. */}
```
waliem, line 50: I think it is really good when you are adding comments to explain what is happening in the code. specially since you often have quite different from other projects. :)

code/src/components/TodoList/TodoListItem.js
```diff
@@ -0,0 +1,99 @@
...
+        <FontAwesomeIcon
+          style={{ cursor: "pointer" }}
+          icon={faTrash}
+          onClick={() => dispatch(todo.actions.removeTodo({ id: item.id }))}
```
waliem, line 54: I'm not sure but isn't it maybe usually preferred to declare and set up this kind of function before the return? I understand why you do it though!

code/src/components/style/colors.js
```diff
@@ -0,0 +1,2 @@
+export const mainColor = "#d98a60";
+export const grayColor = "#d2d3db";
```
waliem, line 2: Nice way of handling repetitive color usage!

waliem (commented):
Well done with this project Sabrina, it is a very impressive work, you have so many functionalities and its very pleasant to use and play around with. I can clearly see that you put a lot of work into it! 👏 

I did not find that much to really point out, I think there is just a couple of things that I think would have been great. First is that even though this project obviously is aimed more at mobile use, I would have suggested you to made the desktop version a bit narrower in its "body", since a lot of the things are fixated to the sides and corners for example it becomes a lot of white space in the middle that doesn't really makes sense. so perhaps put a width or max-width and center the UI to the middle could be a good option. 

Another thing is that I would have loved more comments to your code that explains what is happening, it is not always clear for an outsider (specially for a fellow student) and since you are doing quite differently it would be easier to understand what's going on if there were more comments. I also couldn't really make sense of or locate the scrollToView-part you had, so that could be great to be explained further. The rest where just small pointers, great job, lot of inspiration! 😊

PR #249, opened 2021-11-21
HedvigM todo

HedvigM:
# Project Todos

In this weeks project we started to make our own reducers and actions in the redux store. 

## The problem

I approaced the weeks project by making one funktion at the time. I started with the input field and expanded the functionalities from there. I had a hard time with the syntax of the store and how to write the functionalities to use the store. I solved it throw a lot of googling and looking at different examples. 

If i had more time i would make a sorting function on the app, so there would be possible to sort by date/ time and the different categories. And also a possibility to remove all the todos with a button. Also a would make a local storage. 

## View it live

https://hedvigs-todo.netlify.app


code/src/App.js
```diff
@@ -1,9 +1,24 @@
...
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <Provider store={store}>
+      <Header />
```
hejmaria, line 19: The header is a nice element, but it isnt needed. It is a bit confusing that it says hedvigs header in the top.

code/src/Components/Footer.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import styled from 'styled-components';
+
+export const Footer = () => {
+  return (
+    <StyledFooter>
+      <h1>
+        Hedvig Mejstedt{' '}
+        <span aria-label="lion" role="img">
+          🦁
```
hejmaria: Go lions!

code/src/Components/TodoList.js
```diff
@@ -0,0 +1,208 @@
...
+  };
+
+  return (
+    <Content>
+      <InputField>
+        <h1>TodoList</h1>
+        <div className="input">
+          <div className="inputBox">
+            <label>ToDo:</label>
+            <br />
```
hejmaria: I believe it is best/better practice to use css instead of break-tag to create space. a display: block or flex-direction: column should give the same result.

code/src/Components/TodoList.js
```diff
@@ -0,0 +1,208 @@
...
+                  Choose a Category
+                </option>
+                <option value="Urgent">Urgent</option>
+                <option value="Not urgent">Not urgent</option>
+                <option value="Fun stuff">Fun stuff</option>
+              </select>
+            </div>
+          </div>
+
+          <InputButton disabled={input.length < 1} onClick={onAddToDo}>
```
hejmaria, line 61: nice to prevent posting empty todos.

code/src/Components/TodoList.js
```diff
@@ -0,0 +1,208 @@
...
+      0 17px 50px 0 rgba(0, 0, 0, 0.19);
+  }
+`;
+
+const TheList = styled.div`
+  background-color: white;
+  display: flex;
+  flex-direction: column;
+
+  /* border: 1px solid black; */
```
hejmaria: could remove this :)

hejmaria (commented):
Geat job Hedvig! You write clean code as always, and it is easy to read and understand. I agree with your readme, a local storage would be a nice function - but you have already built a well functioning app! well done!

PR #262, opened 2021-11-22
Nina Rivera - Todo project

NinaAlejandra:
Link to my deployed project:
https://ninarivera-todo-project.netlify.app/

Unfortunately I was sick during this week, but I managed to get it done and fulfill the blue requirements. 

I had an other repo which wouldn't deploy on Netlify, had to delete and re-clone, which is why there are so few commits and there might also be some code I forgot to change in the new repo.


code/src/App.js
```diff
@@ -1,9 +1,39 @@
...
+});
+
+const persistedStateJSON = localStorage.getItem("todosReduxState");
+let persistedState = {};
+
+if (persistedStateJSON) {
+  persistedState = JSON.parse(persistedStateJSON);
+}
+
+const store = createStore(reducer, persistedState);
```
brunsant, line 22: When adding the local storage, the react devtools stops working on Chrome. To fix this, you need to add: 

```
const store = createStore(
  reducer,
  persistedState,
  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
);
```

code/src/components/Header.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import "./header.css";
+
+const date = new Date();
+const todaysDate = date.toString().slice(0, 10);
```
brunsant, line 5: You could also use moment for the date, to keep it consistent with the TodoList component.

code/src/components/Counter.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { useSelector } from "react-redux";
+import "./counter.css";
+
+const Counter = () => {
+  const items = useSelector((store) => store.todos.items.length);
+  const amount = useSelector(
+    (store) => store.todos.items.filter((item) => item.isComplete).length
```
brunsant, line 8: Nice approach using the length to show the amount of tasks.

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,40 @@
...
+
+  const dispatch = useDispatch();
+
+  const onAddTodo = () => {
+    dispatch(todos.actions.addTodo(input));
+  };
+
+  const resetInput = () => {
+    setInput("");
+  };
```
brunsant, line 18: You could add the setInput as a part of the first const: 

`const onAddTodo = () => {
    dispatch(todos.actions.addTodo(input));
    setInput("")
  };`

code/src/reducers/todos.js
```diff
@@ -0,0 +1,50 @@
...
+
+export default todos;
+
+/*const increaseFuction = () => {
+    setCounter(counter +1);
+}
+
+const decreaseFunction = () => {
+    setCounter(counter -1);
+}*/
```
brunsant, line 50: You can delete these if you're not using it.

code/src/index.css
```diff
@@ -11,3 +11,13 @@ code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New",
     monospace;
 }
+
+.container {
+  height: fit-content;
+  min-height: 800px;
+  justify-content: center;
```
brunsant, line 18: I'm not sure this works with out having a Flexbox component.

code/src/components/addTodo.css
```diff
@@ -0,0 +1,27 @@
+.main-container {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+  width: 320px;
+  height: 100px;
+  margin: 20px auto 20px;
```
brunsant, line 8: Since both values are 20px (top & bottom), you can erase one.

code/src/components/todoList.css
```diff
@@ -0,0 +1,79 @@
...
+  border: none;
+  background-color: transparent;
+  font-size: xxx-large;
+  color: whitesmoke;
+  padding: 25px;
+}
+
+.checkbox-container {
+  padding: 0 3px 15px 3px;
+}
```
brunsant, line 47: On the container, you need to add the` position: relative` so the button gets fully covered by new one you add on top of it.

brunsant (approved):
Great project Nina!

Really liked your background picture and color palette. 
I made some comments on the code, but overall there is almost nothing to change. 
Your page passes the validation texts with only 1 error (to add a label to the form) and with some contrast issues. Because the text is white just like the clouds in the background, it can be tricky to read depending on the screen size.

Amazing job this week!!!

PR #265, opened 2021-11-22
ida Utterström Project todo

Idautterstrom:
Link to my deployed project: https://ida-utterstrom-todo-app.netlify.app


code/src/components/Canban.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+const Canban = () => {
```
annaester, line 3: What is Canban?

code/src/components/Todolist.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const onDeleteTodo = (id) => {
+    dispatch(todos.actions.deleteTodo(id));
+  };
+
+  const onCompleteTodo = (id) => {
+    dispatch(todos.actions.toggleTodo(id));
+  };
+
+  const onDoingTodo = (id) => {
```
annaester, line 23: I'm not sure what it is but your toggles are affecting each other. When you check one of them "done" or "doing" the other one gets checked aswell and they toggle each other. So there is something off with the action in the reducer and how you are using it in the TodoList.. 

I belieive it could be cause your using the exact same funtion in your reducer and you are passing the same info to the function. The done and doing gets have the same id so they toggle each other out I guess..?

annaester (commented):
Hi!
Really nice and clean app. I would have liked to see some more styling but I like what you have done so far. There is something off with your different toggle-functions so you can't really toggle between them correctly.

PR #294, opened 2022-04-14
Mariannes-to-do wk 14

MarianneArdin:
Link to my deployed project: https://mariannes-to-do.netlify.app/


qwer-ctrl:
Good work :) All blue requirements are fill up and working correctly. I like that background color is fade out. Maybe missing responsiveness and also would like that the input field would be empty after every added task so I don't have manually delete earlier text :)

PR #298, opened 2022-04-14
Savannah To Do App

savannah-hayes:
Link to my deployed project:

https://savannah-todo-list.netlify.app/

This week I learned a lot about Redux and how powerful it can be. Also about mutability and immutability, which I think is a complex concept to grasp.

I would really like mutability and immutability explained in more detail. For example I had a specific bug in my code, which I know was because of a mutability problem. Inside my map method where I was mapping all the todos, I could not display a timestamp for the time each todo was created at because the timestamp was being mutated each time a new todo was created.


todos/src/components/TimeStamp.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+import { Time } from "styles";
+
+const timeStamp = (created) => {
+  const dateStamp = created.toDateString().split(' ').slice(1, -1).join(' ');
```
EmmaSprings, line 6: Nice! I was looking at this and realising my date stamp is not working as I thought, will try this!

EmmaSprings (commented):
Great work on your to do list! The site works well as it should and I am impressed by getting the timestamp to work, understanding the mutability/immutability. 
Your code is easy to read and is understandable. I like how you've divided your components too, with the dates and timestamp for example. 
Definitely meets the requirements of both blue and red level. Be proud!

PR #299, opened 2022-04-15
Sofia Ringstedt todo

sofiaringstedt:
https://sofia-ringstedt-task-list.netlify.app/


code/src/Styles.js
```diff
@@ -0,0 +1,123 @@
...
+export const DeleteButton = styled.button`
+  position: absolute;
+  top: 10px;
+  right: 10px;
+  background: none;
+  border: none;
+  cursor: pointer;
+`;
+
+export const TodoNumber = styled.h4`
```
rawisou, line 121: Have you explored [Globalstyle](https://scalablecss.com/styled-components-global-styles/)? 

For styling everything globally like containers, h1, h2 etc, GloablStyle can come in handy. :) For a project of this size, its benefit may not be so obvious so more for for larger projects.

code/src/Styles.js
```diff
@@ -0,0 +1,123 @@
+import styled from 'styled-components';
+
+export const Main = styled.main`
+  display: flex;
+  justify-content: center;
+  background: rgb(255,250,250, 0.5);
+  height: 100vw;
```
rawisou, line 7: I wonder what the 100vw here in height does? I tried changing it in the development tool but doesn't seem like the it was taken.

code/src/Styles.js
```diff
@@ -0,0 +1,123 @@
...
+  opacity: 0;
+
+  + label {
+    position: relative;
+    padding-left: 10px;
+    cursor: pointer;
+    display: inline-block;
+    color: #697482;
+    line-height: 25px;
+    &:before, &:after {
```
rawisou, line 79: Great job with pseudo classes!

code/src/index.css
```diff
@@ -1,3 +1,8 @@
+* {
```
rawisou, line 1: With the[ GlobalStyle](https://scalablecss.com/styled-components-global-styles/), the css file can be removed completely. This one can be in the Global.js instead if you use it in your next project. :)

code/src/components/TodoList.js
```diff
@@ -0,0 +1,53 @@
...
+              <span
+                className='button-span'
+                role="img" 
+                aria-label="delete">
+                ✖️
+              </span>
+            </DeleteButton> 
+        </TodoItem>
+      ))}
+      <TodoNumber>Your task(s): &nbsp; {todoList.length}</TodoNumber>
```
rawisou: If you want it to automatically show 0 task and 1 task without s and then plural for many tasks, you can use conditional rendering. Since it's not long/multiple lines of code and simple logic, the ternary operator may be the best obtion. For example:

```
 {todoList.length === 0 || todoList.length === 1 ? (
              <TodoNumber> Your task: {todoList.length} </TodoNumber>
            ) : (
              <TodoNumber> Your tasks: {todoList.length}  </TodoNumber>
            )}
```

rawisou (commented):
Hi Sofia,

Great job this week. Your code is clean and easy to follow.
Your site works as it should. All the functionalities work.

I have small suggestions about GlobalStyle and conditional rendering that I included in the comments. :)

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,56 @@
...
+      <Form onSubmit={onFormSubmit}>
+        <p>{date}</p>
+          <label>
+          <Input 
+          type="text" 
+          value={inputValue}
+          onChange={(e) => setInputValue(e.target.value)}
+          placeholder="Add new task"
+          />
+          <AddButton type="submit">
```
rawisou, line 45: I notice that users could add empty tasks. There're many ways to go about and prevent this. I think the most common way would be to disable the add task button if the input is empty.

code/src/Styles.js
sofiaringstedt: Very good tip, will look into it! Thanks! :)

code/src/Styles.js
sofiaringstedt: It is very hard to see since my background color is very discreet, but it is so the main covers the entire window height! If you change my main background color to red and then take away the height you will see!

code/src/components/TodoList.js
sofiaringstedt: Will do right away!! Thanks!

code/src/components/AddTodo.js
sofiaringstedt: Will also do this right away!! Thanks!

PR #307, opened 2022-04-15
To-Do Project - Michael

michaelchangdk:
Link to my deployed project:


README.md
```diff
@@ -1,13 +1,13 @@
...
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+The brief was to use Redux and React to create a to-do list app.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+I ended up using React-Router as well, and implementing local storage. I can create multiple lists, toggle them as complete or incomplete, delete the tasks, edit the tasks, or delete the entire project lists. Also added aria labels and tabbing to make the site accessible.
```
SteppbySteph, line 7: I noticed you took the time to make it accessible. Nice work!

code/src/App.js
```diff
@@ -1,9 +1,47 @@
...
+import { colors } from "reducers/colors";
+import { icons } from "reducers/icons";
+import Home from "components/Home";
+import Error from "components/Error";
+import Project from "components/Project";
+import Header from "components/Header";
+import NewProject from "components/NewProject";
+import ErrorBoundary from "components/ErrorBoundary";
+
+const enhancer = compose(persistState());
```
SteppbySteph, line 16: Nice seeing you challenge yourself with adding new tools to your project. Maybe add a small comment for the reviewer  that explains the use of these?

code/src/App.js
```diff
@@ -1,9 +1,47 @@
...
+
+const reducer = combineReducers({
```
SteppbySteph, line 18: We have previously worked with one reducer before but three is new! Way to go to adding more to your project :)

code/src/components/ErrorBoundary.jsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import Error from "./Error";
+
+class ErrorBoundary extends React.Component {
```
SteppbySteph, line 4: Is this the "old" way of writing code in React? Curious to know why you found this easier to use than the modern way.

code/src/components/Home.jsx
```diff
@@ -0,0 +1,16 @@
...
+import { useSelector } from "react-redux";
+import NoProjects from "./NoProjects";
+import ProjectList from "./ProjectList";
+import { Container } from "styledelements/elements";
+
+const Home = () => {
+  const projects = useSelector((store) => store.projects.project);
+  return (
+    <Container>
+      {projects.length > 0 ? <ProjectList /> : <NoProjects />}
```
SteppbySteph, line 11: This is a really nice feature of conditional rendering in the navbar. Making it possible to show all created lists and toggle between them is making the app more user friendly. Very good work!

code/src/components/ProjectCarousel.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+const flickityOptions = {
+  initialIndex: 0,
+  pageDots: false,
+  wrapAround: true,
+  freeScroll: true,
+  // autoPlay: 6000,
+};
+
+const ProjectCarousel = (props) => {
```
SteppbySteph, line 12: Loving the carousel feature on mobile version! It's easier using it there than trying it in the browser.

code/src/components/Task.jsx
```diff
@@ -0,0 +1,261 @@
...
+    setEditable(true);
+  };
+
+  const changingText = (e) => {
+    setNewText(e.currentTarget.textContent);
+  };
+
+  const pasteAsPlainText = (event) => {
+    event.preventDefault();
+    const text = event.clipboardData.getData("text/plain");
```
SteppbySteph, line 66: Line 66 and 67 would be nice to comment what these do to your code. :)

code/src/components/Task.jsx
```diff
@@ -0,0 +1,261 @@
...
+          <Icon src={trashgray} alt="delete task." />
+        </IconButton>
+      </IconWrapper>
+    </TaskWrapper>
+  );
+};
+
+export default Task;
+
+const TaskWrapper = styled.div`
```
SteppbySteph, line 146: Good idea to put the styling after all the logic since this component is very long! Functionality comes first ;)

code/src/reducers/colors.js
```diff
@@ -0,0 +1,21 @@
...
+  { id: 8, hex: "#7B39ED", rgba: "123, 57, 237", name: "purple" },
+  { id: 9, hex: "#666666", rgba: "102, 102, 102", name: "gray" },
+];
+
+export const colors = createSlice({
+  name: "colors",
+  initialState: {
+    colors: colorArrayObj,
+  },
+});
```
SteppbySteph, line 21: Didn't know you could create a slice without actually adding a reducer function to it. You seem to have read up on some more deep diving materials about creating and using reducer. You sure left the crawling stage and entered the running stage quickly! *applauding*

code/src/reducers/projects.js
```diff
@@ -0,0 +1,75 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+export const projects = createSlice({
+  name: "projects",
+  initialState: {
+    project: [],
+  },
+
+  reducers: {
```
SteppbySteph, line 9: So impressed to see all actions the user can do with the app. This results in a long list of reducer actions. Wouldn't it be better to use the immutability approach that can "bring increased performance to your app" (https://redux.js.org/faq/immutable-data#what-are-the-benefits-of-immutability). For example I see that you chose to use the 'push' and 'slice' method (mutability approach). As covered in one of the lecture we could write an 'immutable' code instead. For example you could replace line 19 with `state.project = [...state.project, action.payload];`. This was just an idea. The app still works good as it is. As a user I don't seem to find any bugs or so to it.

SteppbySteph (commented):
Wow Michael. As usual you have put a lot of time and effort in creating the projects we get for each week. Seeing you challenging yourself with techniques and functions not covered in the course is going to make you a g.o.a.t developer, sorry I meant great ;) Since you're adding different packages to your project please do share what these do. Otherwise it's great to see you play around with the code and the design. Keep up the good work!! Happy to learn new things here as code reviewing this. // Steph

code/src/App.js
michaelchangdk: The enhancer is for localstorage which was in the "Common Pitfalls" section of the project's brief.

Error boundaries I don't completely understand myself - I was getting some errors in my console log, and had to read some guides. They basically check for errors being passed into the component as far as I remember, and then can render an error component instead. Please don't quote me on it - I was very confused while working on it, and erased it from my memory as soon as I got it to work 😅

code/src/components/ErrorBoundary.jsx
michaelchangdk: I have no idea - I was following the guide for writing Error boundaries on the React documentation page.

code/src/components/ProjectCarousel.jsx
michaelchangdk: Completely agree! The to-do list is much more mobile-friendly than it is desktop friendly. I did primarily focus on making it a mobile web-app than I did on desktop-friendliness (except for animations that are only visible on desktop).

code/src/components/Task.jsx
michaelchangdk: Ah yes, I see that. This is for when people copy-paste text into the to-do list, that it strips everything away and only returns plain text, instead of formatting, or other strange things that happen when you copy-paste directly into the DOM.

code/src/reducers/colors.js
michaelchangdk: Thank you! This was actually just pure chance that I discovered this. All the components that used the colors or icons, I was creating an array for. I decided to test out if it would work just adding it to the reducers and it did, and it meant I could remove a lot of code from my components, which made me really happy!

code/src/reducers/projects.js
michaelchangdk: Yeah, I remember when Maksy was talking about the difference between mutability and immutability, and that we're only able to use the mutability thanks to immer. At that point (the Wednesday curse I call it), I didn't really have the energy to re-write it as immutable. Originally I had four stores - I was storing my tasks separately from my project lists, and including a project ID with each task. Daniel suggested in town hall that I nest my tasks within the individual projects, so I was already quite pressured at that time, because I had to re-write all my reducers, as well as selectors and dispatchers to achieve just a single reducer that contained both projects and tasks.

michaelchangdk:
> Wow Michael. As usual you have put a lot of time and effort in creating the projects we get for each week. Seeing you challenging yourself with techniques and functions not covered in the course is going to make you a g.o.a.t developer, sorry I meant great ;) Since you're adding different packages to your project please do share what these do. Otherwise it's great to see you play around with the code and the design. Keep up the good work!! Happy to learn new things here as code reviewing this. // Steph

Thank you for your feedback Steph - I will definitely be more mindful about including comments for other reviewers in the future, when I'm doing things that aren't included in the brief, or if I discover something. Re-reading my code and your code review actually makes me think I should also do it for myself - for example the error boundary would have been very good to comment for myself. // Michael

PR #308, opened 2022-04-15
Todo app - Emma Högberg

emmahogberg88:
https://todo2do.netlify.app/


code/src/components/Styling.js
```diff
@@ -0,0 +1,77 @@
+import styled from 'styled-components'
+
+export const ButtonSwitch = styled.button`
```
MarianneArdin, line 3: Great use of styled components, I haven't seen media queries embedded in these before!

MarianneArdin (commented):
Great work Emma! Your app looks good and works both in mobile and desktop. Looks like you fulfil most of the red and black level requirements. I love your use of CSS and that the user can choose between dark or light mode. You are such an inspiration!

PR #333, opened 2022-04-18
Project Todo

rijad90:
Link to my deployed project:

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,25 @@
-import React from 'react'
+import React from "react"
+//we import 3 essential packages, Provider, combineReducers & configureStore
+import { Provider } from "react-redux"
+import { combineReducers, configureStore } from "@reduxjs/toolkit"
+//import styled from "styled-components"
+import todos from "reducers/todos"
+import NewTodo from "components/NewTodo"
+// here we create a "pocket" of a bag
+const reducer = combineReducers({
+  todos: todos.reducer,
+})
+
+
+// here we inject the "pocket" in the "backpack"
+const store = configureStore({
+  reducer,
+})
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <Provider store={store}> {/* here we put the "backpack(store)" on the applications "shoulders". everything that is between the Provider will have access to the our store */}
+      <NewTodo />
+    </Provider>
   )
 }
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,16 +1,22 @@
 import React from "react"
-
 import { Provider } from "react-redux"
 import { combineReducers, configureStore } from "@reduxjs/toolkit"
+import styled from "styled-components"
 
 import todos from "reducers/todos"
 import AddTodos from "components/AddTodo"
 import NewTodo from "components/NewTodo"
 
+const Container = styled.div`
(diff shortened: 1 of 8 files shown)


code/src/components/AddTodo.js
```diff
@@ -0,0 +1,46 @@
...
+import styled from "styled-components"
+import uniqid from "uniqid"
+
+import todos from "reducers/todos"
+
+const Label = styled.label`
+  margin: 10px;
+`
+
+const AddTodos = () => {
```
nadialefebvre, line 12: It would be more consistent if both file and component were called the same: your file is named AddTodo and the component is AddTodos.

code/src/components/NewTodo.js
```diff
@@ -0,0 +1,71 @@
...
+const DeleteButton = styled.button`
+  position: absolute;
+  top: 10px;
+  right: 10px;
+  border: none;
+  background: none;
+  cursor: pointer;
+`
+
+const NewTodo = () => {
```
nadialefebvre, line 33: Again about the naming: maybe it would be clearer what this component is when we read your code if it was called "TodoList" or something like that, since it's not really about a new todo :)

code/src/reducers/todos.js
```diff
@@ -0,0 +1,31 @@
+import { createSlice } from "@reduxjs/toolkit"
+
+const todos = createSlice({
+  name: "todos",
+  initialState: {
+    items: [],
```
nadialefebvre, line 6: Good that you used an empty array (so that you removed the hard-coded code used for developing).

code/src/reducers/todos.js
```diff
@@ -0,0 +1,31 @@
...
+  },
+  reducers: {
+    toggleItem: (store, action) => {
```
nadialefebvre, line 9: Nice and simple reducers :)

code/src/components/AddTodo.js
```diff
@@ -0,0 +1,46 @@
...
+    <form onSubmit={onFormSubmit}>
+      <Label>
+        <input
+          type="text"
+          placeholder="New todo"
+          value={inputValue}
+          onChange={(e) => setInputValue(e.target.value)}
+        />
+      </Label>
+      <button type="submit">Submit</button>
```
nadialefebvre, line 41: It would add some extra if you made it required that an input is there before submitting, so the user can't send an empty task.

code/src/App.js
```diff
@@ -1,9 +1,34 @@
...
+import React from "react"
+import { Provider } from "react-redux"
+import { combineReducers, configureStore } from "@reduxjs/toolkit"
+import styled from "styled-components"
+
+import todos from "reducers/todos"
+import AddTodos from "components/AddTodo"
+import NewTodo from "components/NewTodo"
+
+const Container = styled.div`
```
nadialefebvre, line 10: Great that you used styled components :)

nadialefebvre, line 10: Maybe this `div` should be a `main` instead? It's all about semantic html :)

code/src/components/NewTodo.js
```diff
@@ -0,0 +1,71 @@
...
+  text-align: center;
+`
+
+const DeleteButton = styled.button`
+  position: absolute;
+  top: 10px;
+  right: 10px;
+  border: none;
+  background: none;
+  cursor: pointer;
```
nadialefebvre, line 30: Nice that you changed the cursor! (and it's annoying that it's not the default one and that we always have to change it!)

nadialefebvre (commented):
Good job, Rijad, keep it up :)

PR #334, opened 2022-04-18
Project Todos, Tiina Liukkonen

tiiliu:
Link to my deployed project:


code/src/App.js
```diff
@@ -1,9 +1,50 @@
...
+let persistedState = {}
+
+if (persistedStateJSON) {
+  persistedState = JSON.parse(persistedStateJSON)
+}
+
+// Creating store with initial state
+const store = createStore(reducer, persistedState)
+// const store = configureStore({ reducer })
+
```
kolkri, line 24: You could remove all the commented our lines of code

code/src/App.js
```diff
@@ -1,9 +1,50 @@
...
-    <div>
-      Find me in src/app.js!
-    </div>
+      <Provider store={store}>
+        <BrowserRouter>
+          <Routes>
+            <Route path="/" element={<StartPage />} />
+            <Route path="/taskpage" element={<TaskPage />} />
+            <Route path="*" element={<ErrorPage />} />
+          </Routes>
```
kolkri, line 39: nice detail to add the errorpage

code/src/components/TaskPage.js
```diff
@@ -0,0 +1,62 @@
+import React, { useState } from "react"
```
kolkri: You have nicely separated all the components, so the code stays clear. Next project you could also consider to make folder called "pages" and put there your different pages (in this case start/main/error) and then keep the different components in the component file, so it would be even more clear.

code/src/components/StartPage.js
```diff
@@ -0,0 +1,112 @@
...
+                <StyledButton
+                fontSize="1rem"
+                width="140px"
+                fontSizeDesktop="1.2rem"
+                widthDesktop="180px"
+                disabled={!nameInputValue}
+                borderHover="2px solid black"
+                onClick={onNameSet}
+                >
+                    Let's get started
```
kolkri: You could move the StyledButton styling above like other styled components to keep the code clean.

kolkri (commented):
Super nice job with this project, beautiful and clear design. I added few separate comments below. Overall, well done!

code/src/components/TaskList.js
```diff
@@ -0,0 +1,252 @@
...
+                        <DeleteButton onClick={() => onTaskDelete(task.id)}>
+                                <span role="img" aria-label="emoji">
+                                    ✘
+                                </span>
+                        </DeleteButton>
+                    </TaskItem>
+                ))}
+            </StyledTaskList> 
+        }
+
```
kolkri, line 73: Below you are repeating the same code depending how you want to filter them (all/completed/uncompleted). Instead of this you could set outside of the return with if / else statements that what tasklist will you map over (depending on what filter button you clicked).

code/src/reducers/taskitems.js
```diff
@@ -0,0 +1,89 @@
...
+                    }
+                    return 0
+                })
+            } else if (action.payload === "Uncompleted") {
+            state.taskitems = state.taskitems.slice().sort((a, b) => a.complete - b.complete)
+            } else if (action.payload === "Date") {
+                state.taskitems = state.taskitems.slice().sort((a, b) => a.date - b.date)
+            } else {
+                state.taskitems = [...state.taskitems]
+            }
```
kolkri, line 84: super nice funtionality the sorting thing

code/src/components/TaskPage.js
tiiliu: Nice tip! 😊

code/src/components/StartPage.js
tiiliu: This is actually props passed into the styled button component so I can style it differently in different places 😊

code/src/components/TaskList.js
tiiliu: True, a lot of repetition here, thank you for the advice, will try it out! 🙏😊

PR #337, opened 2022-04-19
Project-Todo-Justine

JustineZwiazek:
Link to my deployed project: https://orderly-by-justine.netlify.app/


code/src/App.js
```diff
@@ -1,9 +1,78 @@
...
+let todoStorage = {};
+if (todoStorageJSON) {
+  todoStorage = JSON.parse(todoStorageJSON);
+}
+
+const store = createStore(reducer, todoStorage);
+// Here store is subscribing to using the the localStorage function
+store.subscribe(() => {
+  localStorage.setItem("todos", JSON.stringify(store.getState()));
+});
```
idanaslund, line 57: I didn't think of having local storage, very impressed!

code/src/globalStyle.js
```diff
@@ -0,0 +1,20 @@
+import { createGlobalStyle } from "styled-components";
+
+const GlobalStyle = createGlobalStyle`
```
idanaslund, line 3: Great work with global style!

idanaslund (approved):
Love the design, it's simple but at the same time I can see you have put a lot of effort in to it! You have achieved blue requirements and also some on the higher levels, really great work! As you said the checkbox doesn't fully work with the background, but I still think it looks good. I can't give you any suggestions because I had the same issue. 

Overall great work with your to do list!

PR #342, opened 2022-04-28
Dorothea's todo app

Dorothea-B:
Link to my deployed project:

https://theas-todos.netlify.app/


code/src/App.js
```diff
@@ -1,9 +1,35 @@
...
+    <Provider store={store}>
+      <Background2 />
+
+      <ContentWrapper>
+      <FormAndList />
+      <Counter />
+      </ContentWrapper>
+    
+ 
+    </Provider>
```
tiiliu, line 31: This is purely details, but maybe the empty rows in between could be taken away here :)

code/src/components/Background2.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+
+
+const Form = () => {
```
tiiliu, line 4: This effect is soooo cool and completely gives a lava lamp effect! Super nice! 😍  The only thing I noticed that is that it's said to be good practice that the name of the file and the function name would be matching. Here your file is called Background2 but the function name is Form, so perhaps changing the const name to match the file name (probably something without numbers though), also for semantical reasons (using the word form might be a bit misleading)?

code/src/components/TodoApp.js
```diff
@@ -0,0 +1,121 @@
...
+    ListWrapper,
+    TaskBar,
+    ButtonRow,
+    Cross
+
+} from 'components/StyleComponents'
+
+
+
+export const FormAndList = () => {
```
tiiliu, line 22: Here also the name is not matching the file name, so maybe consider changing one of them :)

code/src/components/TodoApp.js
```diff
@@ -0,0 +1,121 @@
...
+
+    useEffect(() => { setTaskList(allTasks) }, [allTasks])
+
+
+
+    const taskArray = useSelector((store) => store.tasks.items)
+   
+    const finishedTasks = useSelector((store) => store.tasks.items.filter(t => t.status))
+
+    const unfinishedTasks = useSelector((store) => store.tasks.items.filter(t => !t.status))
```
tiiliu, line 41: Here, I think you could have them without empty rows in between, the readability still is preserved :)

code/src/components/TodoApp.js
```diff
@@ -0,0 +1,121 @@
...
+    const filterTasks = (value) => {
+                if (value === 'finished') {
+                    setTaskList(finishedTasks)
+                } else if (value === 'unfinished') {
+                    setTaskList(unfinishedTasks)
+                } else {
+                    setTaskList(allTasks)
+                }
+            }
+
```
tiiliu, line 70: I really like this function and it's very convenient how you use it to filter the tasks on the button click 🤩 As a small detail, the indentation is maybe a tad far 🙃

code/src/components/TodoApp.js
```diff
@@ -0,0 +1,121 @@
...
+            <ListWrapper>
+
+            {taskList.map((task) => (
+                <TaskBar key={task.id} >
+                    <CheckboxStyled htmlFor={task.id}>{task.title} 
+                    <input 
+                    type="checkbox" 
+                    checked={task.status} 
+                    id={task.id} 
+                    onChange={() => dispatch(tasks.actions.toggleState(task.id))} />
```
tiiliu, line 99: I noticed that by keyboard navigation I wasn't able to reach the checkbox at all so maybe look into that? Perhaps adding a tabindex if needed :)

code/src/components/TodoApp.js
```diff
@@ -0,0 +1,121 @@
...
+                    setTaskList(allTasks)
+                }
+            }
+  
+    return ( 
+        <>
+        <FormWrapper>
+            <InputField onSubmit={onFormSubmit}> 
+
+                <Label htmlFor="newTask">Ceep calm and write it down!</Label>
```
tiiliu, line 76: A very small detail, but in "Ceep calm..." the C maybe is suppose to be K? Or maybe it's a choice and if that's the case, ignore this! 😊

code/src/components/TodoApp.js
```diff
@@ -0,0 +1,121 @@
...
+                id={taskArray.length +1}
+                />
+
+                <Submit className="submit-btn" type="submit">Post task</Submit>
+
+            </InputField>
+        </FormWrapper>
+
+
+            <ListWrapper>
```
tiiliu, line 90: Here also you could have the indentation a little more behind :)

code/src/components/TodoApp.js
```diff
@@ -0,0 +1,121 @@
...
+
+    const taskArray = useSelector((store) => store.tasks.items)
+   
+    const finishedTasks = useSelector((store) => store.tasks.items.filter(t => t.status))
+
+    const unfinishedTasks = useSelector((store) => store.tasks.items.filter(t => !t.status))
+
+
+    const newTaskObject = {
+        id: taskArray.length + 1,
```
tiiliu, line 45: One convenient way of using id's which will be for sure unique is with a function called uniqid. The use of it is very easy, just installing (npm install uniqid), importing it on imports and then using it here instead of the array length as such: id: uniqid()  😊

tiiliu (commented):
Really nice work Dorothea, your app looks super nice (love the colours, obviously 😁). The whole vibe is very relaxing and harmonius 🔮  The functionality works without any problems and you've achieved all blue level requirements and some red level too! Your app looks good in all devices, the only thing I noticed that in the mobile version on my iPhone (width 375px), the text on the bottom listing all tasks, finished and unfinished, they're all very close to each other in the small size. Also noticed that those buttons scale pretty nicely in different screen sizes, but in the small mobile size (375px) the buttons go into two rows and on first row they're smaller and on second row they're bigger. Just as an observation for smaller screens 😊  Also I noticed that you're able to add empty tasks, maybe that's also a nice feature to add to block that? Generally, there were some commented out rows in the stylecomponents and index.css, otherwise I've left some comments throughout the code. But in general, great work this week! 🎊

PR #355, opened 2022-11-05
To do 

CeciliaMichele:
Link to my deployed project:
https://dottedtodo.netlify.app/


code/src/components/TaskList.js
```diff
@@ -0,0 +1,60 @@
...
+        </div>
+
+      </section>
+
+    );
+  }
+  return (
+    <section>
+      {allTasks.map((task, index) => functionForMap(task, index))}
+    </section>
```
fannycardell, line 57: I have tried to look at the code here to figure out how to solve the ticking of the checkboxes... I have a totally different solution to it so it was hard to get it interpreted to yours and I haven't used functionFormap. Maybe there could be something with index and broom, maybe only to use index? But then I saw that you have the allTasks.map below the function. Maybe you can try to put the return on line 54 to line 29, I'm not sure if that works, but then you get the mapping inside the function.

fannycardell (commented):
Hello Cissi!

First thing: I really want to write todos when I see your application! 
Bright and playful so you get motivated for both write things up and check them off! Very, very nice! I also really like the version of counter that you put on top, "only...left" is encouraging=) One styling thing that would make it easier to see the list is to make the opacity higher for the white background to the list and/or add a tone to it. And maybe it it also easier to get the brush in line with the box and "todo"

The delete function works fine with the cute brushes, but the checkboxes is now checking all boxes at once and also uncheck all boxes at once. So you actually already have a "complete all" function, but no specific button for it=) See comment below. If you have a look at Daniels code along again, you might can rearrange the code for that.

Again, I really like the colorful and positive vibe with the application! Fabulous!

PR #356, opened 2022-11-05
Project ToDo | Julia Östedt

juliaostedt:
Link to my deployed project: https://whatcha-doing.netlify.app/


code/src/components/TaskList.js
```diff
@@ -0,0 +1,132 @@
...
+          </StyledTask>
+        )
+      })}
+    </section>
+  )
+}
+
+export default TaskList
+
+const StyledTask = styled.article`
```
miadahlgren, line 58: Here are a lot of styled components! Great that you broke them out into several components but maybe it would be easier to find if they were placed outside the TaskList-component :)

miadahlgren (commented):
Sorry for a very short review, but i had very little to comment on in your code since both the design and code looks great. Clean and easy to understand.  The app works great and the responsiveness is well handled. However, one thing i noticed was that when you write a long task the check-box will be a bit squeezed and change form. Other than that I think the assignment is really well accomplished. If you want to continue working on it in the future  you could add localstore to save the tasks when user refreshes page. :) Great job!

PR #361, opened 2022-11-05
Project todos Maria Westling

marwebdesign:
Link to my deployed project: https://project-todo-maria-westling.netlify.app/


code/src/index.css
```diff
@@ -11,3 +11,26 @@ code {
...
+  cursor: pointer;
+}
+
+#checkbox:checked {
+  text-decoration: line-through;
+  background-color: #DF7861;
+}
+
+#checkbox:checked+label {
+  text-decoration: line-through;
```
malmen237, line 35: Very nice looking checkbox, was so surprised that it wasn't a radiobutton. Also nice touch with the line-through, it makes the list extra readable.

malmen237 (commented):
Very nice looking app!

Don't really have any specific comments to make. It's very clean and simple, both in code and in the app. Nice colors and styling that makes it a pleasure to use.

The only improvement I can see is that it's a bit on the small side when it comes to the mobile-view. Tried it my iphone12 och I felt a bit clumsy because my fingers were so big in contrast to the buttons/input-field/checkboxes.

But all in all a very nice page, good job!

PR #365, opened 2022-11-06
Project ToDo Week 11

miadahlgren:
Link to my deployed project:
https://todo-mia-dahlgren.netlify.app/

It was hard and the code got a bit repetitive but in the end it turned out fine. 


code/src/components/AddNewTask.js
```diff
@@ -0,0 +1,58 @@
...
+            <input
+              id="input-add-task"
+              type="text"
+              value={inputValue}
+              placeholder="Add new task"
+              onChange={(event) => setInputValue(event.target.value)} />
+          </label>
+          <AddButtonStyle>
+
+            <button type="submit"><img src={plus} alt="add item" /></button>
```
juliaostedt, line 21: Suggest adding a disabled function to the button so that you're not able to submit an empty task. I took inspo from Happy Thoughts and added the below:

` const buttonDisabled = newTask.length < 2 || newTask.length > 140`
and then:
`<button type="submit" **disabled={buttonDisabled}**>add item</button>`

code/src/components/Footer.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import styled from 'styled-components';
+import Weather from './Weather';
+
+const Footer = () => {
```
juliaostedt, line 5: Nice touch that you added a footer!

code/src/components/Header.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import styled from 'styled-components';
+import moment from 'moment';
+
+const Header = () => {
+  return (
+    <HeaderStyling>
+      <p>{moment().format('MMMM Do YYYY,  h:mm a')}</p>
```
juliaostedt, line 8: The time stamp doesn't seem to automatically update, perhaps worth checking if another moment-format works better?

code/src/styling/ListStyling.js
```diff
@@ -0,0 +1,53 @@
+import styled from 'styled-components';
+
+const ListStyling = styled.div`
+background-color: white;
+opacity: 0.6;
```
juliaostedt, line 5: With opacity you affect the whole content inside the div making the text and buttons a bit difficult to see. I would suggest using a rgba value on background-color instead. You could read more on that here: https://developer.mozilla.org/en-US/docs/Learn/CSS/Howto/Make_box_transparent

code/src/components/RemoveTask.js
```diff
@@ -0,0 +1,34 @@
...
+      </button>
+    </RemoveButtonStyle>
+  )
+}
+
+export default RemoveTask;
+
+const RemoveButtonStyle = styled.button`
+border: none;
+background: transparent;
```
juliaostedt, line 21: Suggest to add a pointer-cursor to the buttons.

code/src/styling/GlobalStyling.js
```diff
@@ -0,0 +1,71 @@
+import styled from 'styled-components';
+import background from 'assets/header2.jpg';
+
+const GlobalStyling = styled.section`
+display:flex;
+flex-direction: row;
+flex-wrap: wrap;
+background-repeat: no-repeat;
+background-image: url(${background});
+background-size: 100%;
```
juliaostedt, line 10: On my larger screen the backgroud image isn't covering the whole page but instead showing a white border at the bottom. Try using contain or cover here instead of 100% as this only affects the width and sets the height to auto.

code/src/styling/GlobalStyling.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+h2{
+    font-size: 16px;
+
+}
+
+}
+@media (min-width: 401px) and (max-width:999px){
+    background-size: 180%;
```
juliaostedt, line 53: If you adjust the background-size as suggested above, this row should not be needed. :)

code/src/styling/GlobalStyling.js
```diff
@@ -0,0 +1,71 @@
...
+
+h2{
+    font-size: 16px;
+
+}  
+.outer-div {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+    justify-content: center;
```
juliaostedt, line 63: Not sure if you need to repeat these four rows or if it perhaps is enough to only change the max-width after the screen-size. Worth testing!

code/src/styling/GlobalStyling.js
```diff
@@ -0,0 +1,71 @@
+import styled from 'styled-components';
+import background from 'assets/header2.jpg';
+
+const GlobalStyling = styled.section`
+display:flex;
+flex-direction: row;
+flex-wrap: wrap;
+background-repeat: no-repeat;
+background-image: url(${background});
```
juliaostedt, line 9: With the background including multiple objects I'd suggest adding a shadow to the image to make your lists stand out more. Perhaps try something like this and adjust the rgba values as wanted:

` background: linear-gradient(rgba(82, 81, 81, 0.5), rgba(0, 0, 0, 0.5)), url(${background});`

code/src/components/Weather.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+const Weather = () => {
```
juliaostedt, line 3: Can't see that this is being used anywhere but remember you mentioned it during the stand-up. Nice feature to have once you get it to work!

code/src/components/TaskCounterHaveTos.js
```diff
@@ -0,0 +1,28 @@
...
+      <p>{countDone()} out of {items.length} tasks completed</p>
+    </TaskCounterStyling>
+  )
+}
+
+export default TaskCounter
+
+const TaskCounterStyling = styled.p`
+color: white;
+`
```
juliaostedt, line 28: This styling part seems to be over-written by something else, I see the text as black/gray.

code/src/components/Header.js
```diff
@@ -0,0 +1,55 @@
...
+export default Header;
+
+const HeaderStyling = styled.div`
+height: 30vh;
+width: 100%;
+font-size: 48px;
+color: grey;
+display: flex;
+justify-content: flex-end;
+align-items: flex-end;
```
juliaostedt, line 21: Nice usage of flex!

code/src/components/TaskListHaveTos.js
```diff
@@ -0,0 +1,73 @@
...
+
+const TaskListHaveTos = () => {
+  const dispatch = useDispatch();
+  const taskItem = useSelector((store) => store.tasks.items);
+  useEffect(() => {
+    const listFromStorage = JSON.parse(localStorage.getItem('taskItem'));
+    if (listFromStorage) {
+      dispatch(tasks.actions.setAllItems(listFromStorage));
+    }
+  }, [])
```
juliaostedt, line 21: Great that you added the local storage!

juliaostedt (commented):
Super impressive creating four different lists and separating them from each other! 👏 I would suggest taking some time to go through the responsiveness as the mobil and tablet versions seem to be a bit far out to the left.

Would also suggest going through the code to remove old comments once you're done (for example the example sorting array in the tasks.js reducer).

PR #371, opened 2022-11-06
To-do List project - Tara

Tara78:
Link to my deployed project:


karlfredrikholm:
Nicely done Tara! The video in the header is a fun feature, but it takes up a lot of space and moves a lot, which makes it pretty annoying if you were to use the actual app underneath it. Maybe the video/animation could have been on a start screen or something, before you enter the actual to-do list app. 

When you put a new task on the list a space between "I have 4 on my to-do list." and the "All done" button appears and it gets larger for each task you put on the list. And if you instead check a task the space moves to between "I Have done 4 by now" and the "Delete all" button. Those are both great features, but that space thing makes it look a little messy.

Over all the app is working great and the code was also easy to follow along. So good job Tara 👏 !

PS. The link in your readme just takes you to your Netlify profile and not to the actual app. 😅

PR #384, opened 2022-11-06
Project-todos by Younas T

YoTesfam:
https://velvety-praline-b94b68.netlify.app/


code/src/reducers/tasks.js
```diff
@@ -0,0 +1,31 @@
...
+
+const tasks = createSlice({
+  name: 'tasks',
+  initialState: {
+    todos: [
+      { id: 1, title: 'groceries', isDone: false },
+      { id: 2, title: 'pick up kids from day care', isDone: false },
+      { id: 3, title: 'traveling on Friday', isDone: false }
+    ]
+  },
```
Kneedler, line 11: If you don't want to start with an an empty list, here his how i did: 

`const initialState = {
  items: []
};

const todos = createSlice({
  name: 'todos',
  initialState,
  reducers: {
    addToDo: (store, action) => {
      const newToDo = {
        id: uniqid(),
        text: action.payload,
        isComplete: false,
        postedTime: moment().format('MMM Do YY')
      };
      store.items = [newToDo, ...store.items]
    },`

code/src/App.js
```diff
@@ -1,9 +1,23 @@
...
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  );
-}
+    <Provider store={store}>
+      <Header />
+      <tasks />
```
Kneedler, line 19: I'm not sure u need the reducer here. I don't know what it does or if it actually displays something.

code/src/components/TaskList.js
```diff
@@ -0,0 +1,118 @@
...
+  background-color: #409A98;
+  border-radius: 20px;
+  box-shadow: 5px 5px 5px #2f4858;
+  @media (min-width: 668px) and (max-width: 1024px) {
+    width: 40%;
+  }
+  @media (min-width: 1025px) {
+    width: 30%;
+  }
+  `
```
Kneedler, line 73: These could be a bit bigger, like 50% (at least on my screen the list is a bit small, but that might also be just an individual preference)

code/src/components/TaskList.js
```diff
@@ -0,0 +1,118 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
+import React from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import tasks from 'reducers/tasks'
+import styled from 'styled-components'
```
Kneedler, line 5: A tip is to write 'styled-components/macro' then in the Dev-tools you can easier see where the style is instead of only the random numbers and letters

Kneedler (commented):
Good job! It's neat code that is not over complicated and that makes it easy to follow. I like the simple layout and the colors u chose. 
Keep up the good work! 
/Sarah

PR #385, opened 2022-11-06
Linda Malms ToDo-list

malmen237:
Link to my deployed project:

https://sloth-productivity.netlify.app/


code/src/styles/mainStyles.js
```diff
@@ -0,0 +1,101 @@
...
+  tablet: '768px',
+  laptop: '1024px',
+  desktop: '2560px'
+};
+
+export const Devices = {
+  tablet: `(min-width: ${sizes.tablet})`,
+  laptop: `(min-width: ${sizes.laptop})`,
+  desktop: `(min-width: ${sizes.desktop})`
+};
```
marwebdesign, line 17: Wow, this is so smart and elegant! Looks so much nicer when doing media queries.

code/src/components/Clock.js
```diff
@@ -0,0 +1,22 @@
+import React, { useState, useEffect } from 'react';
+
+const Clock = () => {
```
marwebdesign, line 3: Really nice touch with the date and time showing

marwebdesign (commented):
Great job Linda! This is such a nice page with well balanced both colors and fonts, and it looks great both on desktop and mobile. The sloth really gives you the feeling that you can really take things easy when all tasks are done. 

I have a hard time giving any constructive feedback, since I think this is so well done. Maybe, it would have been nice to add a 'remove all done tasks' - button you could click when you mark tasks as done. Also, maybe you would like to have the oldest task on top, so you know which one has been there for the longest time and would be nice to take care of first. But this is only a matter of taste. (: 

Well done!

PR #391, opened 2022-11-06
Thérèse project-todos

Tanmark:
Link to my deployed project: https://incredible-concha-b9b5bd.netlify.app/

My reflections on how this week's project turned out: No

Things I'd like to have clarified or explained in more detail: Redux


code/src/components/Footer.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import styled from 'styled-components/macro'
+
+const Footer = () => {
+  return (
+    <FooterWrapper>
+      <FooterParagraph> Made by Thérèse Ånmark at Technigos bootcamp for Web developers 2022
```
jesshansson: I tend to forget to write these footers, but I think it's good to credit youself 👍

code/src/components/Footer.js
```diff
@@ -0,0 +1,37 @@
...
+      </FooterParagraph>
+    </FooterWrapper>
+  )
+}
+
+export default Footer;
```
jesshansson: I think this is usually placed in the very bottom, just as some kind of common practice (but I guess totally not neccessary)

code/src/reducers/todos.js
```diff
@@ -0,0 +1,49 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+const todoData = [
+  { id: 1, text: 'Watch video on actions & reducers', complete: false },
+  { id: 2, text: 'Follow redux codealong', complete: false },
+  { id: 3, text: 'Fork weekly assignment', complete: false }
+];
+
+let counter = todoData.lenght;
```
jesshansson: Misspelled length (I see you've written length in other places). Been there, done that :)

README.md
```diff
@@ -1,13 +1,10 @@
...
-
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+Use redux to create a ToDo list application. The app should list all tasks - completed or uncompleted. You should be able to mark an uncompleted task as complete (and change it back to uncompleted). Be able to add and remove tasks. The app should show a count of either all todos or all uncompleted todos (or both).
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+I mapped in figma before starting, mapped out the reducer, components and styling
```
jesshansson, line 6: Great that you mapped i figma! I haven't learned to do that yet, but I definitely think it's valuable (and most likely used on many workplaces)

code/src/components/ToDoList.js
```diff
@@ -0,0 +1,76 @@
...
+            </ToDoItem>
+          ))}
+        </Wrapper>
+        <ToDo allTodos={allTodos} incomplete={incomplete} dispatch={dispatch} />
+      </>
+    )
+  } else {
+    return (
+      <Wrapper>
+        <h3>Yay! There are no more task for you to dotted</h3>
```
jesshansson: Super if you get this to work 👍

code/src/reducers/todos.js
```diff
@@ -0,0 +1,49 @@
...
+
+const todos = createSlice({
+  name: 'todos',
+  initialState: {
+    todoData
```
jesshansson: I read on StackOverflow that it might be easier to combine your todoData with the const todo. so you only need to "keep track of" one constant. But maybe that depends on wether or not you want some initialdata, for me I just left it empty (and maybe you wont need the const todoData if you want to do that too).

code/src/components/AddToDos.js
```diff
@@ -0,0 +1,98 @@
...
+  const dispatch = useDispatch();
+
+  /* Prevent form from refreshing whole page when submitted */
+  const onFormSubmit = (event) => {
+    event.preventDefault()
+  };
+
+  /* Adding task to the store */
+  const sendToDoToStore = (item) => {
+    if (item.length > 0) {
```
jesshansson: Does this mean that you can only add a task if there already is one posted?

jesshansson (commented):
I definitely think you have a good start! You have a good base with all the code that I think is needed, so it will probably run smoothly once you get the adding of tasks to work 👍 

Solving of code problems can be done in so many different ways, so take my advices with a pinch of salt! I think that you could shorten down your code, for instance some of the parts of "ToDo" and "ToDoList" might be merged into one component. I think it's easier to map the tasks if you do it in the same file as you import the tasks from the reducer file, makes the road a bit simpler so to say (helps when de-bugging).

It might also be good to add some more comments, both for yourself and for others :) 

Good luck with the remaining parts!

PR #392, opened 2022-11-06
Arnau Monforte Coronas- week 11 todos

Iracuzzi:
Link to my deployed project:

https://week11todos.netlify.app/

### DIFF
--- a/code/src/components/Input.js
+++ b/code/src/components/Input.js
@@ -1,0 +1,58 @@
+/* eslint-disable react/button-has-type */
+import React, { useState } from 'react';
+import { useDispatch } from 'react-redux';
+import styled from 'styled-components';
+import { saveTodo } from 'reducers/todoSlice';
+
+const Input = () => {
+  const [input, setInput] = useState('')
+  const dispatch = useDispatch()
+
+  const addTask = () => {
+    console.log(`Adding ${input}`)
+
+    dispatch(saveTodo({
+      item: input,
+      done: false,
+      id: Date.now()
+    }))
+  }
+
+  return (
+    <TodoBarSection>
+      <InputBar type="text" placeholder="Enter new todo..." value={input} onChange={(event) => setInput(event.target.value)} />
+      <Button onClick={addTask}>Add!</Button>
+    </TodoBarSection>
+  )
+}
+
+export default Input;
+
+const TodoBarSection = styled.div`
+  height: 50px;
+  width: 80%;
+  border-radius: 12px;
+  background-color: white;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`;
+
+const InputBar = styled.input`
+  background: none;
+  border: none;
+  margin-left: 12px;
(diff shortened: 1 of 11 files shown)


code/src/components/Input.js
```diff
@@ -0,0 +1,58 @@
...
+      item: input,
+      done: false,
+      id: Date.now()
+    }))
+  }
+
+  return (
+    <TodoBarSection>
+      <InputBar type="text" placeholder="Enter new todo..." value={input} onChange={(event) => setInput(event.target.value)} />
+      <Button onClick={addTask}>Add!</Button>
```
SofieRydmark, line 24: Right now you are only able to add a task by pressing the button. I think if you put this in a form with and onSubmit you can also press enter on keyboard to make it more easy to add tasks :) also if you add  setInput('') at the end of the function it will clear the text in the input area

code/src/components/TodoItem.js
```diff
@@ -0,0 +1,58 @@
...
+  const dispatch = useDispatch()
+  const handleCheck = () => {
+    dispatch(setCheck(id))
+  }
+  const deleteTheTodo = () => {
+    dispatch(deleteTodo(id))
+  }
+
+  const TimeStampHoursMinutes = () => `${String(new Date(id).getHours()).padStart(2, '0')}:${String(new Date(id).getMinutes()).padStart(2, '0')}`
+  const TimeStampDate = () => `${`${new Date(id).toLocaleDateString('en-us', { day: 'numeric' })} ${new Date(id).toLocaleDateString('en-us', { month: 'short' })}`}`
```
SofieRydmark, line 17: nicely done with this

code/src/components/TodoList.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+  const allDeleted = () => {
+    dispatch(deleteAll())
+  }
+
+  return (
+    <>
+      <div>
+        <button type="button" onClick={allDone}>complete</button>
+        <button type="button" onClick={allDeleted}>delete</button>
```
SofieRydmark, line 21: as you know these are currently not workin, which is too bad :). Not sure why not tho, but for me I had to use it in the component instead of importing it from the reducer. Like dispatch(todos.actions.deleteAll())

code/src/store.js
```diff
@@ -0,0 +1,8 @@
+import { configureStore } from '@reduxjs/toolkit';
+import todoReducer from 'reducers/todoSlice';
+
+export default configureStore({
+  reducer: {
+    todos: todoReducer
+  }
+});
```
SofieRydmark, line 8: this could be added directly in the app.js :) and a tip is to add the localstorage for the tasks to save.

SofieRydmark (approved):
I like the clean design of the app, altho the buttons coudl have a little extra styling. And as you already know the buttons is not working right now but otherwise nice clean code! :) are you using the Globalstyles.js? otherwise it might be better to remove since its empy

PR #404, opened 2023-04-20
To do

IrupePozoGraviz:


code/public/index.html
```diff
@@ -4,16 +4,11 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <!--
-      Notice the use of %PUBLIC_URL% in the tags above.
-      It will be replaced with the URL of the `public` folder during the build.
-      Only files inside the `public` folder can be referenced from the HTML.
+    <meta property="og:title" content="Technigo React To do list by Irupé Pozo Graviz"/>
+        <meta property="og:description" content="A website displaying a todo list."/>
+        <meta property="og:image" content="/landing_page.png"/>
```
mvfrid, line 9: Nice adding og-tags! You can check what they look like here if you haven't already:
https://www.opengraph.xyz/url/https%3A%2F%2Firupes-to-do-app.netlify.app%2F

code/src/components/header.js
```diff
@@ -0,0 +1,34 @@
...
+font-size: 15px;
+font-weight: bold;
+color:  rgb(121, 182, 193);
+text-align: center;
+padding-top:18px;
+padding-bottom: 20px
+`
+
+export const Header = () => {
+  const dateTime = moment().format('dddd, MMM Do')
```
mvfrid, line 24: Nice job on adding the date!

code/src/components/toDoList.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }, [showConfetti]);
+
+  useEffect(() => {
+    if (taskList.length > 0) {
+      setHasTasks(true);
+    } else {
+      setHasTasks(false);
+    }
+  }, [taskList]);
```
mvfrid, line 57: Awesome!

code/src/components/toDoList.js
```diff
@@ -0,0 +1,108 @@
...
+
+  return (
+    <div>
+      {hasTasks && (
+        <>
+          <h2>To-do ({todosTodo.length})</h2>
+
+          {todosTodo.length === 0 && !allDoneClicked && (
+            <h3> Viva! You are all done! 🎉 </h3>
+          )}
```
mvfrid, line 73: Excellent work on the conditional rendering!

mvfrid (commented):
You did such a great job this week! The design looks great and you even did some stretch goals. You go Irro!! It looks like you had fun with this task and I hope you did :)

I love the confetti. Great job on incorporating that one and making it happen when all tasks have been marked complete.

It looked good in all screen sizes I checked! Nice and clean design. And nice and clean code as well!

Some improvement ideas if you agree and feel like it:

When the user tries to add a task without writing anything in the input field, the button is unclickable, which is great. It would be nice to add an alert or similar to let the user know that is why they can't click the button.

The check and delete icons look great, but I would maybe add a hover effect on them to make it even more clear that they are clickable. Maybe just change the opacity a little bit? If you really want to go wild I think it would look nice to have the checkmark for the not-done tasks in another color, to really highlight which ones are done and not.

The delete-all image could be cropped a bit to make the container more condense. Now it has a lot of white space around it which makes it difficult to control.

As I mentioned, it looked good in all sizes. You can however achieve this with less code if you add a max-width on the container and then set the width to 100%. That way, you can make sure it never gets wider than a certain size, It will grow and shrink in a nice way! It's a lot easier to have control over it :)

You did a great job this week and should be proud of yourself! Especially given that you missed the quiz task and had to learn twice as much as the rest of us in a shorter time. Snyggt jobbat!

PR #410, opened 2023-04-21
Nora Terbocs - Project to do

noraterbocs:
deployed link: https://redux-to-do-app-nora.netlify.app/


noraterbocs:
I have fixed the font sizes for phone and for bigger screen because they were too small.

PR #411, opened 2023-04-21
https://wondrous-mandazi-bfc5de.netlify.app/

EmiliaGA:


code/src/App.js
```diff
@@ -1,9 +1,20 @@
...
+  });
+
+  const store = configureStore({ reducer })
+
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <Provider store={store}>
+      <TaskPage />
```
edv-rd, line 17: Good practice to not have a lot of code in App.js

code/src/components/Footer.js
edv-rd, line 1: Empty file!

code/src/components/TaskCounter.js
edv-rd, line 1: Nicely made reusable component with styling

code/src/components/TaskList.js
```diff
@@ -0,0 +1,127 @@
...
+              name="tasks"
+              checked={singleTask.complete}
+              onChange={() => onIsCompleteToggle(singleTask.id)}
+              id={singleTask.id} />
+            <div>{singleTask.text}
+              <StyledTimeStamp>
+                {singleTask.timeStamp.toLowerCase()}
+              </StyledTimeStamp>
+            </div>
+            <StyledDeleteButton type="button" onClick={onRemoveButtonClick}>
```
edv-rd, line 35: I think here you don't pass an ID to the delete-function, maybe that's the problem?

code/src/components/TaskList.js
edv-rd, line 1: I think there is a bug with the delete-button, if you have more than one task in the list and press delete on one of the tasks, it's always the first task that will get deleted no matter what

edv-rd (commented):
Nice job! I really liked the functionality and the styling and I think the code looks good. A lot of reusable components using the redux store, I like that.

I noticed one bug which I commented.

👍

PR #412, opened 2023-04-21
Annikas Do-Do List

fikamate:
Not totally done yet


emiliasaberski (commented):
Everything in your code looks good and is in order. The only thing not working for me is the cursor which I guess is supposed to be an emoji, but for me it's only text in both the code and on the site. Other than that everything works as expected! Great job!

PR #418, opened 2023-04-21
Matildas Todo App

mvfrid:
https://matildas-todos.netlify.app/


SofiaGerdmar (approved):
I liked the styling very much, it's colorful and happy. Nice job adding a lot of styled components. The code is neat, tidy and easy to follow. The app is functional and easy to use, I especially liked the tabs to switch between 'done', 'not done' and 'all'. I also think you played around with different fonts in a really nice way. And the blob is the icing on the cake!

PR #427, opened 2023-04-23
Project Todos - Hannah Ek

Nahnahke:


IrupePozoGraviz (commented):
Really Good Hannah, a project well done and I like your code and love the animations
It works really good in iphone 12 mini.

The only thing I thought about was the x and the +  (in the input section) that they are both kind of the same color its not a big problem when im on desktop but as soon as I was in the smaller device it might be a good Idea to difference some more? 

Hanna you are a really good coder and great work

PR #435, opened 2023-04-23
Project to-do list, week 11

fridanordenlow:


code/src/components/ToDoList.js
```diff
@@ -0,0 +1,67 @@
...
+                <div>
+                  {/* <button className="delete-button" type="button" onClick={() => dispatch(tasks.actions.removeTask(singleTask))}>x</button> */}
+                  <button className="delete-single-task-button" type="button" onClick={() => onDeleteTaskBtnClick(singleTask)}>x</button>
+                </div>
+              </div>
+              <hr />
+            </>
+          )
+        })}
+        <p className="invisible-text">Invisible</p>
```
beemailley: This is a clever solution to making sure the empty space always looks like a notepad.  I wonder if it would still work if you didn't include the word "invisible"  on each row?  Mostly because I'm wondering if screen readers will still pick it up.

code/src/components/CkeckboxToDoList.js
```diff
@@ -0,0 +1,67 @@
+/* eslint-disable max-len */
+import React, { useEffect } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import tasks from 'reducers/tasks';
+
+const ToDoList = () => {
```
beemailley: Do you have two components that are named differently but have the same content?  It looks like bother ToDoList.js and CheckboxToDoList.js have the same content?  Is that right?

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,72 @@
...
+    removeAllTasks: (store) => {
+      store.items = []
+
+      localStorage.setItem('taskList', JSON.stringify(store.items))
+    },
+    toggleComplete: (store, action) => {
+      store.items.forEach((task) => {
+        if (task.id === action.payload.id) {
+          task.isCompleted = !task.isCompleted
+
```
beemailley, line 64: I don't think this reducer has been completed.  There's nothing to set task.isCompeted to true.  This is probably why the completed tasks count never increases in the TaskCount.js section.

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,72 @@
...
+          localStorage.setItem('taskList', JSON.stringify(store.items))
```
beemailley, line 65: SO COOL!  I'll have to take a closer look at how you saved everything to local storage!

beemailley (commented):
Overall, it looks good!  
One other comment that I received from my mentor that I'll share with you is about file/folder organization.  Where you have one folder for "regular" components and one folder for "styles", he recommended having basically individual folders for each page (or page-like thing), and then bundling the files for that page together.  E.g.  You could have a Footer folder and in it you put both the Footer.js and the Footer.css.  I switched my folder structure to something like this and it has made it MUCH easier to find the files when they're grouped together.

PR #437, opened 2023-04-23
The todo - w.11 JL

JohLeo:
https://thetodo-jl.netlify.app


code/src/components/AddNote.js
```diff
@@ -0,0 +1,54 @@
...
+import { AddBtn, CounterP } from './Styling/GlobalStyle';
+
+export const AddNote = () => {
+  const [inputValue, setInputValue] = useState('');
+  const [charCount, setCharCount] = useState(0);
+  const dispatch = useDispatch();
+
+  const capitalizeFirstLetter = (str) => {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+  }
```
jonsjak, line 13: Nice detail!

code/src/components/AddNote.js
```diff
@@ -0,0 +1,54 @@
...
+
+  const onInputChange = (event) => {
+    setInputValue(event.target.value);
+    setCharCount(event.target.value.length);
+  }
+
+  const onFormSubmit = (event) => {
+    event.preventDefault();
+    const newNote = {
+      id: Date.now().toString(),
```
jonsjak, line 23: This id works for this project, but when building larger applications with multiple users there's a risk that two posts get the same id if posted at the same time. There are id-generators like uuid (npm package) that prevents this if you want to have look.

jonsjak (commented):
Nice app! Clean responsive design and it works as it should.
Very well structured, easy to follow with good naming and use of variables, components etc.
Because you assign variable names to the data you pass around the text doesn't get cluttered and repetitive. Good job!
One small comment  on the design: Is the contrast on the buttons a bit hard to see (accessibility)?

PS. See additional comments in the code.

PR #451, opened 2023-04-24
week 11 - edvard

edv-rd:


code/src/components/Task.js
```diff
@@ -0,0 +1,55 @@
...
+font-size: 15px;
+color: var(--foreground-primary-color);
+flex: 1;
+text-align: center;
+transition: color 0.1s ease-out;
+cursor: pointer;
+
+
+
+&.completed {
```
EmmaHoltegaard, line 27: Neat way of adding some conditional styling. Didn't know you could target classnames in styled components like this, but it makes sense once you see it.

code/src/components/TaskList.js
```diff
@@ -0,0 +1,116 @@
...
+    setToggledActionBar(!toggledActionBar);
+  }
+
+  const dispatch = useDispatch();
+
+  const handleCompleteAll = () => {
+    dispatch(tasks.actions.completeAllTasks())
+  }
+  const handleDeleteAll = () => {
+    if (window.confirm('Are you sure you want to delete all?')) {
```
EmmaHoltegaard, line 93: Nice touch adding the window. Adds a lot to the user experience that you can't accidentally just delete everything with a single click.

code/src/reducers/tasks.js
```diff
@@ -0,0 +1,31 @@
...
+  reducers: {
+    addTask(state, action) {
+      state.todos.push(action.payload)
+    },
+    completeTask(state, action) {
+      const todo = state.todos.find((t) => t.id === action.payload)
+      todo.complete = !todo.complete
+    },
+    deleteTask(state, action) {
+      // state = state.filter((t) => t.id !== action.payload)
```
EmmaHoltegaard, line 21: Good job getting around the filter-method (from the live session I got the impression filter() wasn't the best choice for this). Your solution is really nice and not overly complicated (which I love).

EmmaHoltegaard (commented):
Great job, as always. Simple design, but superuser-friendly. Everything seems to work as it should, only minor problem I noticed, is that the input field doesn't clear after you submit a task. Otherwise, everything looks good and your code is clean and easy to read.

PR #455, opened 2023-04-28
Pull req so that a code review can be performed, code is faulty and the browser is blank atm

SandraMadeleine:
Site is now fixed and will be handed in, will not make a new pull request.


EmiliaGA (commented):
Code looks neat and tidy. Can't see any obvious problems but can't view the site as there is no Netlify link. Hopefully you'll get some help with what's the matter. Keep up the good job :)

PR #458, opened 2023-05-02
Project Todos

LinaAdamsson:


LinaAdamsson:
For this week I wanted to design an app that emphasized the feeling of accomplishment when a task is completed, hence the huge checkboxes and the sound effect when ticking them. Not done at all, will be fun to continue working on it.

